Issue #12, opened 2018-11-28
Multi-cursor support


nblackburn:
At the moment when are presented with suggestions, upon selecting one, only the first selection is populated. It would be nice if all selections were populated in the cases where you are using multiple cursors.

ciaoben:
+1

![tabnine bug](https://user-images.githubusercontent.com/4526061/49992753-39c8e300-ff85-11e8-841d-1f0d70ab6040.gif)

PR #18, opened 2018-12-14
Fix multi cursor

jackfranklin:
Hey,

Thanks firstly for the tool + this plugin! I'm such a fan I made a license purchase last week :)

#12 was really frustrating me so I took a go at fixing it. Rather than use `args.position`, this uses `editor.selections` instead. It seems to do the trick based on my testing :)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,7 +22,7 @@ export function activate(context: vscode.ExtensionContext) {
   vscode.commands.registerTextEditorCommand(
     COMMAND_NAME,
     (editor, edit, args: CommandArgs) => {
-      editor.selections.forEach((selection: vscode.Selection) => {
+      editor.selections = editor.selections.map((selection: vscode.Selection) => {
         edit.insert(
           selection.active.translate(0, -args.old_prefix.length),
           args.new_prefix
@@ -35,6 +35,9 @@ export function activate(context: vscode.ExtensionContext) {
           selection.active.translate(0, -args.old_prefix.length),
           selection.active.translate(0, args.old_suffix.length),
         ));
+
+        let new_position = selection.active.translate(0, -args.new_suffix.length);
+        return new vscode.Selection(new_position, new_position)
       })
     }
   );
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,20 +22,20 @@ export function activate(context: vscode.ExtensionContext) {
   vscode.commands.registerTextEditorCommand(
     COMMAND_NAME,
     (editor, edit, args: CommandArgs) => {
-      edit.insert(
-        args.position.translate(0, -args.old_prefix.length),
-        args.new_prefix
-      );
-      edit.insert(
-        args.position.translate(0, args.old_suffix.length),
-        args.new_suffix
-      );
-      edit.delete(new vscode.Range(
-        args.position.translate(0, -args.old_prefix.length),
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -22,20 +22,20 @@ export function activate(context: vscode.ExtensionContext) {
...
-      edit.insert(
-        args.position.translate(0, args.old_suffix.length),
-        args.new_suffix
-      );
-      edit.delete(new vscode.Range(
-        args.position.translate(0, -args.old_prefix.length),
-        args.position.translate(0, args.old_suffix.length),
-      ));
-      let new_position = args.position.translate(0, -args.old_prefix.length);
-      editor.selection = new vscode.Selection(new_position, new_position);
```
jackfranklin: I have removed the code that sets the new position and the `editor.selection`, only because from my testing I didn't seem to need it for the cursor to behave as expected.

I'm happy to add this back though and map `editor.selections` to be the new selections based on `new_position`.

zxqfl:
Thank you for the PR! I will get this merged as soon as I have time to test it.

One case to test is: if `old_suffix` is empty but `new_suffix` is not, is the cursor placed correctly? You can test this by creating a file with:

```
foo(a)
foo(b)
f
```
TabNine should suggest `foo(` and the suggestion will have `old_suffix == ""`, `new_suffix == ")"`.

jackfranklin:
Ah good point! I think I will need to bring the positioning logic back. I'll try to update the pr when I get a moment.

jackfranklin:
@zxqfl I think you were right that we needed some logic. https://github.com/zxqfl/tabnine-vscode/pull/18/commits/d2f06b26d1493b7f556ba14c5aa7b8b1327e6c16 adds it.

This means that in this case:

```
foo(a)
foo(b)
f [invoke tabnine]
```

Tab nine suggests `foo(`, with a `new_suffix` of `)`, and the cursor ends up in the middle of the brackets, which I believe is correct:

```
foo([CURSOR HERE])
```

If you could check + merge + release it that'd be great, thank you!

zxqfl:
Did you do any configuration to get VS Code to allow autocompletion suggestions with multiple selections? When I try to test it, I don't see any autocompletion popup.

jackfranklin:
@zxqfl nope, just works out the box for me. Could you try doing that on the master branch and seeing if that works? That way we'll know if I've broken it or if it's your VSCode!

zxqfl:
It also doesn't work for me on master. Are we running the same VS Code? I'm running VS Code 1.30 on OSX.

jackfranklin:
Likewise - I've tested this branch on OSX and Ubuntu and it's worked on both. I'm just digging through my settings to see if anything relating to autocomplete appears.

jackfranklin:
@zxqfl could you try clearing out your settings.json file (or commenting it all out) and see then?

jackfranklin:
For clarity, this is what I see (excuse the squashed up resolution, but you get the idea!): https://cl.ly/00b57ef0e8dc

jackfranklin:
Are you using vscode vim? I've just discovered a bug in that which breaks multi cursors.

zxqfl:
Yes I am! Thanks for looking into it. Sorry I didn't get back to you earlier. I will get this merged this weekend.

zxqfl:
Thanks again!

jackfranklin:
@zxqfl thank you!

PR #22, opened 2019-01-12
Update to TabNine 1.0.10

AlexanderEkdahl:
Thank you for this amazing tool.

I used the `dl_binaries.sh` script to download the latest version of the binaries and ran the extension using the provided `./vscode/launch.json`. Everything appears to be working and `tabnine::version` reports `1.0.10`

* Does this justify the minor version bump? from `1.1.0` to `1.2.0`?
* Is there any remaining work to do?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 **/out
 
 binaries/*
-!binaries/1.0.7
+!binaries/1.0.10
 
 registration_key
 auth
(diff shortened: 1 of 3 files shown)


zxqfl:
Hi Alexander,

Thanks for the PR, but I will not merge it. TabNine will automatically update the binary once it is installed. For this reason, I prefer not to update the binaries packaged with the plugins unless I want to upgrade the protocol.

AlexanderEkdahl:
That makes sense! I misunderstood how it worked.

Issue #2, opened 2018-10-21
appears to break snippets support completely


ericblade:
I haven't historically used snippets a lot, but I was going through all my extensions and realized that I had a few snippet extensions that I could've been using, but really haven't been.  I went to try them out, and noticed that snippet support was not working at all.  Uninstalling tab nine fixed it, and all my snippets work now.

ericblade:
Sure, I have a few of my own snippets defined in my global javascript.json snippets, those did not work (even with snippets set to be on top), also have "ES6 Mocha Snippets" installed . . I think I removed all the others that I had previously installed.. so, that's what I currently have.  No snippets showed up with TabNine installed, they do after uninstalling.

vojto:
@zxqfl Experiencing same issue.

I have a JS snippet with trigger `log`, which will expand into `console.log(...)`.

Now if I type `log`, TabNine gives me suggestions, and when I press Tab, first of those suggestions is used.

This is not what I want – I wanna use my snippet with trigger `log`.

langolf:
Emmet abbreviations in css/scss don't show at all. For html-developer it's better to have emmet in high priority in rules (for speed) and tabnine in classnames/mixins/variables. I think it will be good to have settings for tabnine how much line it can take and after it will go all editor suggestions. And of course priority list.
![db2a404e-fef5-4148-87dc-b5b5e2cb0e0a](https://user-images.githubusercontent.com/1225205/48870880-63ee1180-edda-11e8-961d-f6f776389b92.jpg)
![32529c82-8c13-4c58-bdad-d5e30cb7977a](https://user-images.githubusercontent.com/1225205/48870881-63ee1180-edda-11e8-8267-fa0cf7ba4616.jpg)

jasonycw:
C# code also show popup with an empty box as well
![image](https://user-images.githubusercontent.com/4518597/62358148-b32e3980-b546-11e9-9b88-763cc03954bb.png)

NoobTW:
Can I only enable TabNine in some languages or disable TabNine in some specific languages?

I found TabNine really useful when I'm writing JS. But It kills emmet abbreviations in HTML and CSS files.

j6k4m8:
@NoobTW and for others:

You can disable TabNine by setting 
```
"tabnine.disable_file_regex": [
        "*.html"
    ],
```

in your settings.json file. Make sure that `"emmet.triggerExpansionOnTab": true` is also included to ensure that Emmet continues to work in HTML files.

[...]

PR #48, opened 2019-08-30
fix: #2

xream:
I am not familiar with vscode plugin development. But I think I fixed this issue.

Thank you for your hard work.


boaz-codota:
@dimacodota ??

PR #72, opened 2020-02-12
fix cursor position in case of suggested suffix

dimacodota:
- Substitute command was removed - the purpose of this command was to handle the cursor in case of new suffix suggested by tabnine, replacement of the old prefix and suffix, and position the cursor before the suggested new suffix for all selections. instead, [SnippetString](https://code.visualstudio.com/api/references/vscode-api#SnippetString) usage was added to the competition item creation (`makeCompletionItem`), which allows to add a literal template snippet that can handle the cursor position (`appendTabstop`) between the inserted text values .
 - package-lock was updated due to deprecate usage of the flatmap-stream library (error while npm install : Not Found: flatmap-stream@https://registry.npmjs.org/flatmap-stream/-/flatmap-stream-0.1.0.tgz)

**after this fix:**

p4EjWM3ra8
LJcOS2Vmy1

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,53 +15,8 @@ import {Mutex} from 'await-semaphore';
 const CHAR_LIMIT = 100000;
 const MAX_NUM_RESULTS = 5;
 const DEFAULT_DETAIL = "TabNine";
-const COMMAND_NAME = "TabNine Substitute";
 
 export function activate(context: vscode.ExtensionContext) {
-
-  vscode.commands.registerTextEditorCommand(
-    COMMAND_NAME,
-    (editor, edit, args: CommandArgs) => {
-      editor.selections = editor.selections.map((selection: vscode.Selection) => {
-        const editorPrefix = editor.document.getText(new vscode.Range(selection.active.translate(0, -args.old_prefix.length), selection.active));
-        if (args.old_prefix != editorPrefix) {
-          return selection;
-        }
-        let oldPrefix;
-        let newPrefix;
-        if (args.old_prefix.length > 0) {
-          oldPrefix = args.old_prefix;
-          newPrefix = args.new_prefix;
-        } else {
-          const tail = editor.document.getText(new vscode.Range(editor.document.lineAt(selection.active.line).range.start, selection.active.translate(0, -args.old_prefix.length)));
-          oldPrefix = tail;
-          newPrefix = tail + args.new_prefix;
-        }
-        const newSuffix = args.new_suffix;
-        const editorSuffix = editor.document.getText(new vscode.Range(selection.active, selection.active.translate(0, args.old_suffix.length)));
-        let oldSuffix: string;
-        if (args.old_suffix == editorSuffix) {
-          oldSuffix = args.old_suffix;
-        } else {
-          oldSuffix = "";
-        }
-        edit.insert(
-          selection.active.translate(0, -oldPrefix.length),
-          newPrefix
-        );
-        edit.insert(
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -15,54 +15,9 @@ import {Mutex} from 'await-semaphore';
 const CHAR_LIMIT = 100000;
 const MAX_NUM_RESULTS = 5;
 const DEFAULT_DETAIL = "TabNine";
-const COMMAND_NAME = "TabNine Substitute";
 
 export function activate(context: vscode.ExtensionContext) {
 
-  vscode.commands.registerTextEditorCommand(
```
dimacodota: this command was removed,  [SnippetString](https://code.visualstudio.com/api/references/vscode-api#SnippetString) was used instead

src/extension.ts
```diff
@@ -265,13 +211,6 @@ export function activate(context: vscode.ExtensionContext) {
   }
 }
 
-interface CommandArgs {
```
dimacodota: unused interface removed

src/extension.ts
```diff
@@ -185,22 +140,13 @@ export function activate(context: vscode.ExtensionContext) {
...
-      title: "accept completion",
-    };
-    item.insertText = args.entry.new_prefix;
-    item.range = new vscode.Range(args.position.translate(0, -args.old_prefix.length), args.position);
+    
+    item.insertText = new vscode.SnippetString(args.entry.new_prefix)
+      .appendTabstop()
+      .appendText(args.entry.new_suffix);
+
+    item.range = new vscode.Range(args.position.translate(0, -args.old_prefix.length), args.position.translate(0, args.entry.old_suffix.length));
```
dimacodota: the range was changed from the begging of the old prefix (`args.old_prefix.length`) to old suffix (`args.entry.old_suffix.length`)

src/extension.ts
```diff
@@ -185,22 +140,13 @@ export function activate(context: vscode.ExtensionContext) {
...
-    item.command = {
-      arguments: [arg],
-      command: COMMAND_NAME,
-      title: "accept completion",
-    };
-    item.insertText = args.entry.new_prefix;
-    item.range = new vscode.Range(args.position.translate(0, -args.old_prefix.length), args.position);
+    
+    item.insertText = new vscode.SnippetString(args.entry.new_prefix)
+      .appendTabstop()
```
dimacodota: appendTabstop() positions the cursor after the new prefix and before new suffix

src/extension.ts
```diff
@@ -185,22 +140,13 @@ export function activate(context: vscode.ExtensionContext) {
     : vscode.CompletionItem
   {
     let item = new vscode.CompletionItem(args.entry.new_prefix);
-    item.insertText = "";
```
dimacodota: duplicate initialization was removed

dimacodota:
> LGTM. Seems to be quite a hefty change, we should test for regression

agree

src/extension.ts
zxqfl: One thing to check here is that the text after the cursor is what we expect it to be. The most likely way for it to not be what we expect is if the user has multiple selections, and the text after the cursor is different in the multiple selections. Here's an example of a case where we should make sure not to replace the `..` after the second `foo`:
image

dimacodota: hey Jacob, thanks for bringing this to my attention, this is actually the range property behaviour, it operates only on the main selection

multiselct-cursor position

dimacodota:
@avichay77 FYI

PR #73, opened 2020-02-16
keep the precedence of the vscode auto-import completion items

dimacodota:
-  set the sortText prefix to U+FFFF9 + the current index, in order keep the precedence of the auto-import completion items.
vscode uses U+FFFF5 as sorting string for auto imported completion items
https://github.com/microsoft/vscode/blob/6ec6f9e3f4d17f0cef0480688a1f350b9141e567/extensions/typescript-language-features/src/features/completions.ts#L56 

-  init the `kind` property in case of undefined ( language service is not started )

-  remove the `preselect` property and let vscode to select the matching item

auto-import

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -17,6 +17,9 @@ const MAX_NUM_RESULTS = 5;
 const DEFAULT_DETAIL = "TabNine";
 const COMMAND_NAME = "TabNine Substitute";
 
+// vscode uses U+FFFF5 as sorting string for all auto imported completion items
+// (https://github.com/microsoft/vscode/blob/6ec6f9e3f4d17f0cef0480688a1f350b9141e567/extensions/typescript-language-features/src/features/completions.ts#L56)
+const SORT_TEXT_PREFIX = '\uffff9';  
 export function activate(context: vscode.ExtensionContext) {
 
   vscode.commands.registerTextEditorCommand(
@@ -186,7 +189,7 @@ export function activate(context: vscode.ExtensionContext) {
   {
     let item = new vscode.CompletionItem(args.entry.new_prefix);
     item.insertText = "";
-    item.sortText = new Array(args.index + 2).join('0');
+    item.sortText = SORT_TEXT_PREFIX + args.index;
     item.range = new vscode.Range(args.position, args.position);
     let arg: CommandArgs = {
       old_prefix: '',
@@ -213,8 +216,8 @@ export function activate(context: vscode.ExtensionContext) {
     } else {
       item.detail = args.detailMessage;
     }
-    item.preselect = (args.index === 0);
-    item.kind = args.entry.kind;
+
+    item.kind = args.entry.kind || vscode.CompletionItemKind.Snippet;
     return item;
   }
 


vladfrangu:
I've tested this locally (by modifying the out file) and it seems to work as intended, any chance we can get a pair of 👀 at this? I also see a conflict right now..

boaz-codota:
@dimacodota ??

vladfrangu:
Why was this closed?

boaz-codota:
This was resolved in another PR. Why? Are you experiencing any issues regarding it?

vladfrangu:
Oh, it was? I didn't catch that, but it seems to not be respected still, unless I'm understanding it wrong, I'd expect TS suggestions first 👀 

image

boaz-codota:
@vladfrangu yea you are understanding it incorrectly. You should expect TabNine suggestions 1st always.

boaz-codota:
We currently do not support configuration for that (nor do we plan to), but we would love to receive any ideas regarding how to improve our suggestions and overall experience.

vladfrangu:
> @vladfrangu yea you are understanding it incorrectly. You should expect TabNine suggestions 1st always.

But..why? The point of this PR was to change exactly that, and make TS suggestions come before TabNines (the GIF attached in the PR description shows just that), as it gives far more accurate results context wise... I love TabNine, I really do, but personally, I'd expect TS suggestions first, and as I start typing out what I'm looking for, making the query return less and less, see both TS and TabNine directly after so I can go "Down arrow, tab" and get the sweet AI autocomplete.

This can also be a very simple config option to add, a simple true/false which adds the prefix from this PR if enabled, or doesn't otherwise 👀 

Even from the screenshot I shared, the TabNine autocompletes are really..not..helpful (in this one example). Of course, this is just feedback, I still love how helpful TabNine is overall, and it has helped me skip a lot of time wasted on tedious things!

boaz-codota:
This PR didn't intent to address what you describe.

I totally understand your expected behaviour, and will let @avichay77 know so we will take this into account. 

Please do not expect that behaviour to change currently, as we strongly believe that setting a lower priority to TabNine is not a sustainable solution. But we are experimenting with different ways to improve its behaviour.

Thanks you for caring and using TabNine! We highly appreciate your feedback!

vladfrangu:
This PR actually addressed what I described, here's a screenshot with the code and it's result:

Before "this PR" (before applying the prefix manually)
image

After "this PR" (after applying the prefix manually, and disabling preselect like in this PR) (the highlighted client in the second image is the TabNine suggestion)
image
image

---

Of course, this wouldn't be something you'd want enforced (by all means, for most people, TabNine is probably what they want to see first when it comes to suggestions, and from a product stand point it makes a LOT of sense), but rather something that a user can choose to change (so you can optionally change it if you so desire). I can also respect not wanting to add this behavior (although it would really help me and others out, as we want TabNine to *augment* our suggestions, not overtake them entirely (or require us to scroll down through suggestions to get to the ones provided either by TS or extensions that provide Intellisense based on the language one uses); I know this has deferred one of our teammates from using TabNine :c), hence this is just a feedback! I'm glad to hear that it will be taken into account, and I wish all of you a great day! ❤️

PR #81, opened 2020-03-14
escape $ sign that is preserved for tab stops in snippet strings and remove the tabstop if there is no suffix

dimacodota:
- escape $ sign that is preserved for tab stops in snippet strings.
- remove the tabstop if there is no suffix - otherwise it will mark the text after the inserted snippet as selected.

**first issue:**

escape-tab-stop

**second issue:**

selected-text

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -179,7 +179,7 @@ export function activate(context: vscode.ExtensionContext) {
     }
   }
   function escapeTabStopSign(value){
-    return value.replace("$","\\$");
+    return value.replace(new RegExp("\\$", 'g'), "\\$");
   }
 
   function isMarkdownStringSpec(x: any): x is MarkdownStringSpec {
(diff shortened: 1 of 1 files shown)


dimacodota:
@yahave - FYI

dimacodota:
@amircodota - FYI

src/extension.ts
```diff
@@ -175,6 +178,9 @@ export function activate(context: vscode.ExtensionContext) {
       return documentation;
     }
   }
+  function escapeTabStopSign(value){
```
amircodota: Why is the $ sign special? Are there other characters that are special and that need t be escaped too?

dimacodota: $ defines interpolation expression, there are other special characters like {} but they are relevant only in context of $ sign, e.g. : 

- ${expression} - special,  interpolation  that evaluates
- $1 - special, tab stop 
- {} - not special, not evaluates

PR #92, opened 2020-04-06
tabnine config as vs code command

dimacodota:
config-page

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -217,7 +217,7 @@ export function activate(context: vscode.ExtensionContext) {
       }
       if (new RegExp(r).test(line)) {
         return false;
-      }``
+      }
     }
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -18,6 +18,22 @@ const DEFAULT_DETAIL = "TabNine";
 
 export function activate(context: vscode.ExtensionContext) {
 
+  const command = 'TabNine::config';
+  const commandHandler = () => {
+    const request = tabNine.request("1.0.7", {
```
amircodota, line 23: We have a builtin command in the protocol for that. You can simply send

```
            "Configuration": {
            }
```

dimacodota: done

PR #94, opened 2020-04-12
bugfix/completions-stop-when-process-exits-while-waiting-for-a-response

amircodota:
… to the response

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -298,11 +298,11 @@ class TabNine {
       }
     });
 
-    let timeout = new Promise((resolve, reject) => {
+    let timeout = new Promise((_resolve, reject) => {
       setTimeout(() => reject('request timed out'), 1000);
     });
 
-    let procExit = new Promise((resolve, reject) => {
+    let procExit = new Promise((_resolve, reject) => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -271,31 +271,71 @@ class TabNine {
...
         }
-        this.rl.once('line', (response) => {
+        const onResponse: (input: any) => void = (response) => {
           let any_response: any = JSON.parse(response.toString());
           resolve(any_response);
-        });
+        };
+        this.rl.once('line', onResponse);
+
+        unregisterFunctions.push(() => this.rl.removeListener('line', onResponse));
```
dimacodota, line 292: do we need to unregister events that were register with `once` ?

amircodota: Yes, if the promise was resolved or rejected, and these were not fired yet, we need to unregister or we are going to leak.
Specifically, the process does not exit under normal circumstances, so if we just register for it with evey request, we are going to leak

dimacodota: 👍

dimacodota: nice catch...

PR #98, opened 2020-05-03
Issue/auto imports

dimacodota:
this implementation consist of:

- calling the vscode.executeCodeActionProvider which will provide if there are suggested code actions for selected range of text (same as the lightbulb).
- execute the suggested "import" code action if any

auto-import


src/extension.ts
```diff
@@ -19,6 +20,26 @@ export function activate(context: vscode.ExtensionContext) {
       "Configuration": {}
     });
   };
+  const importsCommand = async (editor, edit, {completion}) => {
+    setTimeout(async () => {
+      try {
+        let selection = editor.selection;
+        let completionSelection = new vscode.Selection(selection.active.translate(0, -completion.length), selection.active);
+        let codeActionCommands = await vscode.commands.executeCommand<vscode.CodeAction[]>('vscode.executeCodeActionProvider', editor.document.uri, completionSelection, vscode.CodeActionKind.QuickFix);
+        let importCommands = codeActionCommands.filter(c => c.title.toLocaleLowerCase().includes("import"));
```
nimrod-codota, line 29: Haven't run this code, but is it maybe possible that the `kind` of the `CodeAction` is more specific than `QuickFix`, allowing us to filter it by the `kind` rather than by title? If its possible I think that would be preferable. I Can't think of another code action which will include 'import' but if we can be more explicit with `kind` filtering, that sounds better to me.

src/extension.ts
```diff
@@ -19,6 +20,26 @@ export function activate(context: vscode.ExtensionContext) {
...
+        if (importCommands.length) {
+          let [firstCommand] = importCommands;
+          await vscode.workspace.applyEdit(firstCommand.edit);
+          await vscode.commands.executeCommand(firstCommand.command.command, firstCommand.command.arguments);
+          await vscode.commands.executeCommand(COMPLETION_IMPORTS, {completion} );
+        }
+      } catch (error) {
+        console.log(error);
+      }
+    }, 400);
```
nimrod-codota, line 39: I'm not entirely sure why we're waiting here, i'm guessing VSCode will only add the quickfix only after the line is actually added. I'm wondering if there's a way for us to grab the selection before we `setTimeout`, I'm afraid this can lead to a race condition

src/extension.ts
dimacodota: all the`QuickFix`s (add missing imports, add single import, add existing , etc. ) have the same kind of `vscode.CodeActionKind.QuickFix` and i already filtered them to match only the `vscode.CodeActionKind.QuickFix` by passing  the `kind` to `executeCodeActionProvider` command (the last argument)  

- `Import [module]`
- `add [module] to existing import`
(hence the lowercase check)
maybe we should explicitly filter for those two (with regex or something), WDYT?

src/extension.ts
dimacodota: the problem here is with the `vscode.executeCodeActionProvider` command that is not returning any suggestion if invoked with less than 200ms, but we definitely can grab the selection before the timeout, good catch, i'll change it 👍

src/extension.ts
nimrod-codota: I'm in for documenting this, and I think that using regexes is a really good idea to do so in code, I vote for it. As far as I remember its also pretty performant in JS compared to string ops (indexOf etc.)

dimacodota: done 👍

PR #100, opened 2020-05-12
adding client-metadata

Eransho:
Adding the new `client-metadata` argument

### DIFF
--- a/src/TabNine.ts
+++ b/src/TabNine.ts
@@ -96,7 +96,7 @@ export class TabNine {
       "--client=vscode",
       "--client-metadata",
       "clientVersion=" + vscode.version,
-      "pluginVersion=" + ext.getVersion(),
+      "pluginVersion=" + ext.version,
       ...additionalArgs
     ];
     const binary_root = path.join(__dirname, "..", "binaries");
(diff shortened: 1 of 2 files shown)


src/TabNine.ts
```diff
@@ -195,5 +203,29 @@ export class TabNine {
...
+}
+
+export class TabNineExtension {
+  private extension: vscode.Extension<any>;
+  
+  private constructor(ext: vscode.Extension<any>) {
+    this.extension = ext;
+  }
+
+  getExtensionPath(): string {
```
dimacodota, line 215: you can defined them as getters: `get version() {}` and then call them as properties `ext.version`

Eransho:
BTW, is there a reason not to use `vscode.env.appName` as the client string argument instead of the fixed "vscode" string?

src/TabNine.ts
```diff
@@ -87,8 +91,12 @@ export class TabNine {
   }
 
   private static runTabNine(inheritStdio : boolean = false, additionalArgs: string[] = []): child_process.ChildProcess {
+    const ext = TabNineExtension.getInstance();
```
dimacodota, line 94: maybe we should create this only once? not for each binary start, this is "static" metadata related to the extension that will not change between the binary instances

dimacodota:
> BTW, is there a reason not to use `vscode.env.appName` as the client string argument instead of the fixed "vscode" string?

it will behave differently in vscode-insiders, you will get "Visual Studio Code - Insiders" while in vscode you will get "Visual Studio Code"

Eransho:
> > BTW, is there a reason not to use `vscode.env.appName` as the client string argument instead of the fixed "vscode" string?
> 
> it will behave differently in vscode-insiders, you will get "Visual Studio Code - Insiders" while in vscode you will get "Visual Studio Code"

I assumed it will be a different value. The question is do we want to distinguish between the "different" products

src/TabNine.ts
Eransho: cool, thanks 👍

src/TabNine.ts
Eransho: makes sense

src/TabNine.ts
```diff
@@ -87,8 +91,12 @@ export class TabNine {
...
     const args = [
       "--client=vscode",
+      "--client-metadata",
+      "clientVersion=" + vscode.version,
+      "pluginVersion=" + ext.version,
```
amircodota, line 99: @dimacodota maybe we also one if this is an insiders build or not?

dimacodota: we will get this as part of the `vscode.version`, if it is insiders the version will be `-insiders`

PR #107, opened 2020-06-05
support import of default exports 

dimacodota:
```
Button.ts

const Button = () => <button>hello</button>
export default Button

```
```
other.ts

<Button>
```
default-imports

### DIFF
--- a/src/importsHandler.ts
+++ b/src/importsHandler.ts
@@ -2,6 +2,9 @@ import { Selection, commands, CodeAction, workspace, TextEditor, CodeActionKind 
 
 const importStatement = /Import ([\S]*) from module [\S]*/;
 const existingImportStatement = /Add ([\S]*) to existing import declaration from [\S]*/;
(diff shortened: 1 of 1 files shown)


src/importsHandler.ts
```diff
@@ -43,6 +46,6 @@ function filterSameImportFromDifferentModules(importCommands: CodeAction[]): Cod
 }
 
 function getImportName({ title }) {
-    let statement = title.match(importStatement) || title.match(existingImportStatement);
+    let statement = importStatements.map(statement => title.match(statement)).find(Boolean);
```
nimrod-codota, line 49: `find(Boolean` gets the only element from the list? maybe its a good idea to ensure its a single one? just a thought..

dimacodota: find by definition will return single value (first)if exists. the usage of the Boolean function will return the truthy value, it is logically equal to `.filter(a => a)` IMO if you are familiar with the convention it is more readable. WDYT?

PR #119, opened 2020-07-07
handle the request timeout for getCapabilities request

dimacodota:
handle the request timeout for getCapabilities request


src/extension.ts
```diff
@@ -208,7 +208,7 @@ export async function activate(context: vscode.ExtensionContext) {
       }
     }
     item.preselect = (args.index === 0);
-    item.kind = args.entry.kind || vscode.CompletionItemKind.Module;
```
dimacodota, line 211: should not part of this fix, reverted from master

PR #120, opened 2020-07-07
Issue/get capabilities on focus

dimacodota:


src/extension.ts
```diff
@@ -31,11 +31,24 @@ const initHandlers = once((tabNine: TabNine, context: vscode.ExtensionContext) =
...
+        });
+      }
+    });
+  }
   const tabNineExtensionContext =  getContext();
   const tabNine = new TabNine(tabNineExtensionContext);
-  const { enabled_features } = await tabNine.getCapabilities();
-  const isCapability = (capability) => enabled_features.includes(capability)
-
+  const { enabled_features } =  await getCapabilitiesOnFocus();
```
amircodota, line 50: Won't that cause activate to block until the window is focused? Is that advisable?

dimacodota: not exactly, the activate function is async as well, so it will not block the vscode extension host activation but will block this specific activation function until vscode will receive a focus

amircodota: But doesn't that mean that vscode considers that extension not activated yet, isn't that a problem? For example, they might have a timeout for extensions to activate properly, and de-activate them if timeout elapsed.

Maybe we should let the activate function complete and do it in the background?

dimacodota: the extension will be activated as the activate function was called and return a promise, you can see it in the [vscode API](https://github.com/microsoft/vscode/blob/3c77120d88b57b302e86c84cd65e33be7e8433e9/src/vs/workbench/api/common/extHostExtensionService.ts#L417)

so in action vscode will call the activate function will gat a promise because of `async` and will mark it as activated, and the extension itself will be handle on focus (in background)

dimacodota: the other option is to wrap all the activation function in `new Promise` and return it, but it would be equivalent to the `async` statement correct me if i'm wrong

amircodota: Maybe move the whole code to the `onDidChangeWindowState` and not wait for it in `activate`?

src/extension.ts
```diff
@@ -30,236 +30,250 @@ const initHandlers = once((tabNine: TabNine, context: vscode.ExtensionContext) =
...
+          }
+        });
+      }
+    });
+  }
+  const tabNineExtensionContext = getContext();
   const tabNine = new TabNine(tabNineExtensionContext);
-  const { enabled_features } = await tabNine.getCapabilities();
-  const isCapability = (capability) => enabled_features.includes(capability)
+  getCapabilitiesOnFocus().then(({ enabled_features }) => {
```
dimacodota, line 50: this was changed

src/extension.ts
dimacodota: @amircodota  - i've changed it to `then` instead of `async`, everything else is formatting please review again, thanks

PR #127, opened 2020-07-20
Task/use deactivate

dimacodota:
use deactivate as binary request

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -21,10 +21,11 @@ const CHAR_LIMIT = 100000;
 const MAX_NUM_RESULTS = 5;
 
 const DEFAULT_DETAIL = "TabNine";
+let tabNine: TabNine = null;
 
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -21,10 +21,11 @@ const CHAR_LIMIT = 100000;
 const MAX_NUM_RESULTS = 5;
 
 const DEFAULT_DETAIL = "TabNine";
+let tabNine: TabNine = null;
```
amircodota, line 24: why?

src/extension.ts
```diff
@@ -21,10 +21,11 @@ const CHAR_LIMIT = 100000;
...
 
 export function activate(context: vscode.ExtensionContext) {
   const tabNineExtensionContext = getContext();
-  const tabNine = new TabNine(tabNineExtensionContext);
```
amircodota, line 27: why?

PR #129, opened 2020-07-23
close stdouts on deactivate

dimacodota:


amircodota:
Is this the right solution?

Not all plugins call de-activate.
Also, the editor might abruptly without invoking de-activate.

I think the right solution is to handle the error of writing to stdout and exit in an orderly fashion, without panicing

dimacodota:
> Is this the right solution?
> 
> Not all plugins call de-activate.
> Also, the editor might abruptly without invoking de-activate.
> 
> I think the right solution is to handle the error of writing to stdout and exit in an orderly fashion, without panicing

you right, will handle in the binary. closing it

PR #138, opened 2020-09-08
Add no-lsp flag

boaz-codota:


boaz-codota:
WIP: DO NOT MERGE BEFORE THE `no-lsp` flag version is out and is updated!!!!

svipas:
@boaz-codota Out of curiosity why this flag will be added and set to true?

boaz-codota:
TabNine allows users to turn on "semantics completions" using the `tabnine::sem` command. Basically, they are the same suggestions the LSP gives you if you have one. On vscode & intellij almost everyone has LSP suggestions, so there is no need for TabNine to even try and suggest LSP suggestions (this often resolves to duplicate suggestions that just bombard TabNine users).
This flag tells TabNine's binary to not provide such suggestions.
I hope that answers your question @svipas

svipas:
@boaz-codota Wow, nice! But if I have disabled semantic completion it would be good to have `no-lsp=false` instead of `no-lsp=true`? Or what you mean is in all cases it's better to use `no-lsp=true` because almost every language have LSP ant it will reduce duplicate suggestions. I'm just thinking if this affects suggestions in a bad way if I will have semantic completions disabled and `no-lsp=true`.

boaz-codota:
@svipas semantic completions are useless in intellij & vscode, as there are awesome plugins that fulfill that need, so need for us to supply them. Therefor, this change is a push for disabling semantic completions in those IDEs.

Your spinoff plugin would preserve his current behaviour, no change there if this is your concern.

svipas:
@boaz-codota Yeah I know, I'm just curious about the changes you're doing because TabNine is awesome and I'm using it every day. Thanks for letting me know, I'm very happy with these changes. TabNine is improving every single day and that is awesome to know. So as far as I understand semantic completions will be gone in the future?

boaz-codota:
@svipas currently there is no plan to do such thing. Semantics are untouched as they have value for some IDEs (sublime/vim).
And if you like sublime, there are few experiments I'm quite excited about that are being launched as we speak, so keep tuned!

dimacodota (approved):
LGTM

dimacodota (approved):
LGTM

PR #144, opened 2020-10-11
Add prettier

boaz-codota:


dimacodota (approved):
LGTM, 2 questions:
why double quotes and not single?
why explicit parentheses for a single parameter?

boaz-codota:
> LGTM, 2 questions:
> why double quotes and not single?

@dimacodota I prefer defaults over flavour above all.

> why explicit parentheses for a single parameter?

(Again defaults, but also:) This saves time when adding more parameters, and it is better matched to TypeScript (that requires the parentheses to insert a type).

PR #147, opened 2020-10-14
Extended validatorSelection analytics

shakedbr-codota:


src/validator/ValidatorHandlers.ts
```diff
@@ -63,13 +65,14 @@ function eventDataOf(
   currentSuggestion: Completion,
   allSuggestions: Completion[],
   reference: string,
-  threshold: number
+  threshold: number,
+  ignore: boolean
```
dimacodota, line 69: set it to default false

dimacodota (approved):
LGTM, see my comment

src/validator/ValidatorHandlers.ts
shakedbr-codota: Done.

PR #149, opened 2020-10-19
Refactor & Bump minimum required VSCode version

boaz-codota:


src/TabNine.ts
```diff
@@ -1,300 +1,124 @@
...
+    } finally {
+      console.log("Binary request failed.");
+      release();
+    }
   }
 
-  private isChildAlive(): boolean {
-    return this.proc && !this.childDead;
-  }
+  private readLineWithLimit(timeout: number): Promise<any> {
```
boaz-codota, line 56: This makes the whole old mechanism unneeded. I just wrapped it with an exception, and it just works.

src/TabNine.ts
```diff
@@ -1,300 +1,124 @@
...
-        return value;
-      },
-      (err) => {
-        unregister();
-        throw err;
-      }
-    );
+      return JSON.parse(result.toString());
+    } finally {
+      console.log("Binary request failed.");
```
dimacodota, line 51: why it always "Binary request failed" ?

src/extensionContext.ts
```diff
@@ -3,6 +3,8 @@ import { TabNineExtensionContext } from "./TabNineExtensionContext";
 
 const EXTENSION_SUBSTRING = "tabnine-vscode";
 
+export const tabnineContext: TabNineExtensionContext = getContext();
+
 export function getContext(): TabNineExtensionContext {
```
dimacodota, line 8: remove the export

src/capabilities.ts
```diff
@@ -1,28 +1,31 @@
 import { TabNine } from "./TabNine";
```
dimacodota, line 1: remove

src/commandsHandler.ts
```diff
@@ -1,48 +1,35 @@
-import { API_VERSION, TabNine, StatePayload, StateType } from "./TabNine";
+import { TabNine } from "./TabNine";
```
dimacodota, line 1: remove

src/TabNine.ts
```diff
@@ -1,300 +1,124 @@
...
+import { createInterface, ReadLine } from "readline";
+import BinaryRun, { binaryRunInstance } from "./BinaryRun";
+import {
+  API_VERSION,
+  CONSECUTIVE_RESTART_THRESHOLD,
+  MAX_SLEEP_TIME_BETWEEN_ATTEMPTS,
+  SLEEP_TIME_BETWEEN_ATTEMPTS,
+} from "./consts";
 
 export class TabNine {
```
dimacodota, line 12: do not export it if it is became internal

dimacodota (commented):
see my comments

src/BinaryRun.ts
```diff
@@ -0,0 +1,42 @@
+import * as child_process from "child_process";
+import BinaryVersionFetcher from "./BinaryVersionFetcher";
+import BinaryPaths from "./BinaryPaths";
+import { tabnineContext } from "./extensionContext";
+
+export default class BinaryRun {
+  constructor(private binaryVersionFetcher: BinaryVersionFetcher) {}
```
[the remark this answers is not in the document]
boaz-codota: Well I kind of went for the OOP implementation. I have no problem with the non OOP way of doing this. Changing now.

src/TabNine.ts
```diff
@@ -1,300 +1,124 @@
...
-      `vscode-remote=${context?.isRemote}`,
-      `vscode-remote-name=${context?.remoteName}`,
-      `vscode-extension-kind=${context?.extensionKind}`,
-      ...additionalArgs,
-    ].filter(Boolean);
-    const binary_root = path.join(__dirname, "..", "binaries");
-    const command = TabNine.getBinaryPath(binary_root);
-    return child_process.spawn(command, args, {
-      stdio: inheritStdio ? "inherit" : "pipe",
+      this.rl.once("line", resolve);
```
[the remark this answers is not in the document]
boaz-codota: This is the thing: there is no need of unregistering the listeners. If the promise resolves with a timeout, the `rl.once(line, resolve)` will just resolve without doing anything (because the promise is already fulfilled).

src/TabNine.ts
boaz-codota: Was like this, didn't bother to change it

src/extensionContext.ts
boaz-codota: Can't it is used in few places

src/TabNine.ts
boaz-codota: It is not internal

src/TabNine.ts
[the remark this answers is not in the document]
dimacodota: yes, this is correct. however, the concern here is not with promise suddenly being resolved after it was rejected. but that the EventEmitter listener array will endlessly grow, causing a memory leak. 
see the [implementation](https://github.com/nodejs/node-v0.x-archive/blob/ed0d1c384cd4578f7168633c838f1252bddb260e/lib/events.js#L190-L209) of once, where the listener is cleaned only when `on` is invoked

src/extensionContext.ts
dimacodota: and why not use the `tabnineContext` ?

src/TabNine.ts
dimacodota: is it used for typing only?

src/validator/ValidatorClient.ts
```diff
@@ -117,25 +118,22 @@ async function request(
...
-      }, timeToSleep);
-    }),
-  ];
-  return Promise.race(promises);
+
+  return new Promise((resolve, reject) => {
+    const id = getNanoSecTime();
+
+    validationProcess
+      .post({ ...body, id, version: VALIDATOR_API_VERSION }, id)
```
dimacodota, line 130: nice 👍

src/validator/ValidatorClient.ts
```diff
@@ -109,6 +109,7 @@ export interface Completion {
 }
 
 let validationProcess: ValidatorProcess = null;
+
 async function request(
```
dimacodota, line 113: you do not need it to be async anymore

src/validator/ValidatorClient.ts
```diff
@@ -241,18 +237,18 @@ class ValidatorProcess {
     this.restartChild();
   }
 
-  async post(any_request: any, id: number): Promise<Promise<any>> {
+  async post(any_request: any, id: number): Promise<any> {
```
dimacodota, line 240: ah, yeah, good one, `Promise<Promise<any>>` makes no sense...

src/TabNine.ts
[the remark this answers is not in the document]
boaz-codota: It doesn't matter as the GC takes care of that. It will eventually either resolve or the process will die, so we will create a new process and free the pointer to the old process. Then the GC should take care of the rest...

src/capabilities.ts
```diff
@@ -1,13 +1,18 @@
...
-export const VALIDATOR_CAPABILITY = "vscode.validator";
+export enum Capability {
+  ON_BOARDING_CAPABILITY = "vscode.onboarding",
+  VALIDATOR_CAPABILITY = "vscode.validator",
+}
+
+let isCapability: (c: Capability) => boolean;
+
+export function isCapabilityEnabled(c: Capability): boolean {
+  return isCapability ? isCapability(c) : false;
```
dimacodota, line 12: Elvis operator?  `isCapability?.(c)`

src/capabilities.ts
```diff
@@ -1,13 +1,18 @@
...
+}
 
-export function getCapabilitiesOnFocus(): Promise<{
-  isCapability: (string) => boolean;
-}> {
+export function fetchCapabilitiesOnFocus(): Promise<void> {
```
dimacodota, line 15: why do you think this kind of API is better? it forces a temporal coupling between fetchCapabilitiesOnFocus and isCapabilityEnabled, you must call the first in order it to operate properly

src/extension.ts
```diff
@@ -1,57 +1,51 @@
...
+    },
+    ...COMPLETION_TRIGGERS
+  );
+
+  if (isCapabilityEnabled(Capability.VALIDATOR_CAPABILITY)) {
+    initValidator(context);
+  }
+
+  if (
+    process.env.NODE_ENV !== "production" ||
```
dimacodota: extract it to isDevMode or something

src/binary/Binary.ts
```diff
@@ -0,0 +1,113 @@
...
+export default class Binary {
+  private consecutiveRestarts: number = 0;
+  private isRestarting: boolean = false;
+  private mutex: Mutex = new Mutex();
+
+  private proc?: child_process.ChildProcess = null;
+  private rl: ReadLine;
+
+  constructor() {
+    setImmediate(() => {
```
dimacodota, line 21: why to do async stuff in the ctor?

dimacodota (commented):
comments

dimacodota (approved):
LGTM

src/capabilities.ts
boaz-codota: Yep, it goes for a global thing instead of propogating this function all the way down.

src/binary/Binary.ts
boaz-codota: Well originally to not block the initialisation. But I removed it.

PR #174, opened 2020-10-27
Add Typescript Strict Mode

boaz-codota:
Also compile typescript to es2018 due to the fact that we target Node 10.x or higher. See: https://stackoverflow.com/a/57607634

### DIFF
--- a/src/binary/requests.ts
+++ b/src/binary/requests.ts
@@ -35,7 +35,7 @@ export function autocomplete(requestData: {
   region_includes_beginning: boolean;
   region_includes_end: boolean;
   max_num_results: number;
-}): Promise<AutocompleteResult> {
+}): Promise<AutocompleteResult | undefined | null> {
   return tabNineProcess.request({
     Autocomplete: requestData,
   });
@@ -43,7 +43,7 @@ export function autocomplete(requestData: {
 
 export function configuration(
   body: { quiet?: boolean } = {}
-): Promise<{ message: string }> {
+): Promise<{ message: string } | null | undefined> {
   return tabNineProcess.request(
     {
       Configuration: body,
@@ -52,11 +52,13 @@ export function configuration(
   );
 }
 
-export function setState(state) {
+export function setState(state: any) {
   return tabNineProcess.request({ SetState: { state_type: state } });
 }
 
-export function getState(content: Record<any, any> = {}) {
+export function getState(
+  content: Record<any, any> = {}
+): Promise<State | null | undefined> {
   return tabNineProcess.request<State>({ State: content });
 }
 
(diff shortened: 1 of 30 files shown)


src/binary/requests.ts
```diff
@@ -52,11 +52,13 @@ export function configuration(
   );
 }
 
-export function setState(state) {
+export function setState(state: any) {
```
dimacodota, line 55: why any?

src/handleUninstall.ts
```diff
@@ -5,15 +5,15 @@ import { uninstalling } from "./binary/requests";
...
     const uninstalledPath = path.join(extensionsPath, ".obsolete");
     const isFileExists = (curr: fs.Stats, prev: fs.Stats) => curr.size != 0;
     const isModified = (curr: fs.Stats, prev: fs.Stats) =>
       new Date(curr.mtimeMs) >= new Date(prev.atimeMs);
-    const isUpdating = (files) =>
-      files.filter((f) =>
-        f.toLowerCase().includes(tabnineContext.id.toLowerCase())
-      ).length != 1;
+    const isUpdating = (files: string[]) =>
+      files.some((f) =>
```
dimacodota, line 14: this is a bug - in normal case it will have one file named by the `id` for example tabnine-1.0.0
in case of an update, it should get more than one files prefixed with the `id` for example tabnine-1.0.1

so  some will always return true

src/selectionHandler.ts
```diff
@@ -127,10 +129,10 @@ async function handleImports(editor: TextEditor, completion: any) {
         completionSelection,
         CodeActionKind.QuickFix
       );
-      let importCommand = findImports(codeActionCommands)[0];
+      let importCommand = findImports(codeActionCommands!)[0];
 
-      if (importCommand) {
-        await workspace.applyEdit(importCommand.edit);
+      if (importCommand.edit) {
```
dimacodota, line 134: are you sure this is correct, what would be returned from findImports in case of no import command?

dimacodota (commented):
see my questions

src/binary/requests.ts
boaz-codota: Because I have no idea what types it accepts. You know?

dimacodota: https://github.com/codota/acpl-paid/blob/791760a8a98a2823e717b71b16bb32d42983d69f/tabnine/server/src/api/v2_1.rs#L73

boaz-codota: Fucking shit 🤦🏻‍♂️

dimacodota (approved):
LGTM

PR #199, opened 2020-11-10
Notifications with tests

boaz-codota:


dimacodota (approved):
LGTM

src/notifications/pollNotifications.ts
```diff
@@ -0,0 +1,39 @@
...
+    return;
+  }
+
+  notifications.notifications.forEach(
+    (notification) => void handleNotification(notification)
+  );
+}
+
+async function handleNotification(notification: Notification): Promise<void> {
+  return vscode.window
```
dimacodota, line 30: let's add a `once` wrapper on top of it with a notification.id as a key

dimacodota (approved):
see my comment

src/notifications/pollNotifications.ts
boaz-codota: Why? That sounds like a hack that will bite us in the ass in the future

PR #210, opened 2020-11-15
Shorter `displayName`

sclickk:
Fixed since the current `displayName` spans the width of the entire screen when opening the right-click menu on the status bar. I also moved the programming language names into `keywords`.

image


mukundshah:
@boaz-codota 

>  We try hardly to increase the reach of Tabnine to developers who have yet to try it. Having such long title improve our presence in VSCode Marketplace, and hence support the cause.

Aren't the description and keywords meant for such purpose??
Seriously, who else is doing such quirky stuff?

PR #224, opened 2020-11-22
upon alpha install, store the installed version in globalState and …

dimacodota:
…check the updates against it. this was added because vscode manages the version only in package.json and does not override it with the vsix name. so when installing ex-123.vsix with package version 3 vscode will install it as  ex-1


boaz-codota (approved):
🚀

PR #230, opened 2020-11-24
promotion status element added, notification shown + status shown bin…

dimacodota:
promotion status element added, notification shown + status shown binary requests


boaz-codota (approved):
LGTM

PR #248, opened 2020-12-06
Feature/vanilla keyword origin

shaiahr:
This branch adds support for the new VANILLA_KEYWORD origin. 
The change should affect only statistics events so that we can now differentiate between local index suggestions (a.k.a vanilla) and keywords from the keywords file (a.k.a vanilla_keyword).
See Jira task requirements here:
https://codota.atlassian.net/browse/DEV-1343


boaz-codota (approved):
LGTM

PR #251, opened 2020-12-06
clear alpha state on uninstall + show alpha version in status tooltip

dimacodota:
clear alpha state on uninstall + show alpha version in status tooltip


boaz-codota (approved):
LGTM

PR #256, opened 2020-12-07
set state once for the same text

dimacodota:
set state once for the same text

### DIFF
--- a/src/statusBar/stusBarActionHandler.ts
+++ b/src/statusBar/stusBarActionHandler.ts
@@ -10,7 +10,7 @@ import {
   StatePayload,
   STATUS_BAR_NOTIFICATION_PERIOD,
 } from "../consts";
-import { getStatusBarData, resetDefaultStatuses, setPromotionStatus } from "./statusBar";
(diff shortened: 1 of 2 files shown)


src/statusBar/stusBarActionHandler.ts
```diff
@@ -21,9 +21,11 @@ export default async function handleStatus(
 ): Promise<void> {
   registerStatusHandlingCommand(status, context);
 
-  void setState({
-    [StatePayload.STATUS_SHOWN]: { text: status.message },
-  });
+  if (getStatusBarData()?.text !== status.message){
```
boaz-codota, line 24: Instead of exposing the statusBar (leaky abstraction) can we use something like `statusBarTextIs(status.message)`?

boaz-codota (approved):
Minor suggestion, otherwise LGTM

src/statusBar/stusBarActionHandler.ts
dimacodota: 👍

PR #268, opened 2020-12-20
support action via status and notifications API

dimacodota:


src/binary/requests/statusBar.ts
```diff
@@ -1,20 +1,17 @@
...
 
 export type StatusBarStatus = {
     id: string;
     message: string;
     options: {
-      key: string;
-      action: StatusBarActionActions;
-    }[];
-    notification_type: string;
+      action: MessageActions;
```
boaz-codota, line 12: Haven't we removed the options kinun?

src/notifications/executeNotificationAction.ts
```diff
@@ -1,18 +1,15 @@
...
 ): Promise<void> {
-  const selectedAction = notification.options.find(
-    ({ key }) => key === selected
-  );
 
-  switch (selectedAction?.action) {
-    case NotificationActions.NONE:
+  switch (selectedAction) {
+    case MessageActions.OPEN_HUB:
+      return openConfigWithSource(StateType.NOTIFICATION)();
```
boaz-codota, line 11: I did not implement that part in intellij

boaz-codota (changes requested):
Notice the `action` thing

PR #309, opened 2021-01-24
Publish vsix to open vsx

bilucodota:


.github/workflows/tag.yml
```diff
@@ -52,4 +52,8 @@ jobs:
     - name: Publish
       run: npm run vsce:publish -- ${{ steps.package-version.outputs.current-version }}
       env:
-        VSCE_PAT: ${{ secrets.VSCE_PAT }}
\ No newline at end of file
+        VSCE_PAT: ${{ secrets.VSCE_PAT }}
+    - name: Publish To Open VSX
```
dimacodota, line 56: maybe mark it with `continue-on-error`?  otherwise, it will fail the whole job

dimacodota (approved):
had a small suggestion, besides this LGTM

PR #316, opened 2021-01-28
add new StartupActions request and execute actions on startup

yonip23:


src/binary/startupActionsHandler.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default async function executeStartupActions(): Promise<void> {
+    await sleep(BINARY_STARTUP_GRACE);
+    const actionsResult = await getStartupActions();
+
+    if (actionsResult?.actions.includes(MessageActions.OPEN_HUB)){
+        return openConfigWithSource(StateType.STARTUP)();
+    }
+
+    return Promise.resolve();
```
boaz-codota, line 14: redundant, delete line (you are using async so by default the promise resolves

boaz-codota (approved):
You the man

dimacodota (approved):
til

src/binary/startupActionsHandler.ts
yonip23: IJ doesnt like it if i remove this line, plus i copied this code from `executeNotificationAction.ts`:

```typescript
export default async function executeNotificationAction(
  selectedActions: MessageActions[] | undefined
): Promise<void> {

  if (selectedActions?.includes(MessageActions.OPEN_HUB)){
    return openConfigWithSource(StateType.NOTIFICATION)();
  }
  return Promise.resolve();
}
```

yonip23: image

PR #439, opened 2021-03-24
enrich plugin events

yonip23:
example of added fields:
```json
{
   "timestamp":"1616591659536",
   "os":"linux-Ubuntu-x64",
   "kernel":"5.8.0-45-generic",
   "cpu":"Intel®-Core™ i7-10750H",
   "cores":"12",
   "speedGHz":"2.6",
   "memoryBytes":"33406607360"
}
```

PR #445, opened 2021-03-29
Enrich plugin analytics

yonip23:
A follow up to https://github.com/codota/tabnine-vscode/pull/439


src/reports/reporter.ts
```diff
@@ -11,24 +13,38 @@ export enum EventName {
...
   context.subscriptions.push(reporter);
 }
 
 export function report(event: EventName): void {
-  reporter.sendTelemetryEvent(event);
+  if (inTestMode) return;
+
+  void (async () => {
+    const data = await getReportData();
+    console.log(JSON.stringify(data));
```
bilucodota, line 35: `console.log`

src/reports/reportData.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return specsCache;
+}
+
+export default async function getReportData(): Promise<ReportData> {
+  const specs = await getSpecsCache();
+
+  return {
+    timestamp: `${new Date().getTime()}`,
+    os: `${specs.os.platform}-${specs.os.distro}-${specs.os.arch}`,
```
boaz-codota, line 78: why not split it? why join it? It will be harder to slice and dice this way.

src/reports/reportData.ts
```diff
@@ -0,0 +1,85 @@
...
+    kernel: `${specs.os.kernel}`,
+    cpu: `${specs.cpu.manufacturer}-${specs.cpu.brand}`,
```
boaz-codota, line 80: same

src/reports/reporter.ts
```diff
@@ -11,24 +13,37 @@ export enum EventName {
...
 }
 
 export function report(event: EventName): void {
-  reporter.sendTelemetryEvent(event);
+  if (inTestMode) return;
+
+  void (async () => {
+    const data = await getReportData();
+    reporter.sendTelemetryEvent(event, data);
+  })();
```
boaz-codota, line 36: Would be easier to read this without the await
```typescript
getReportData().then(data => reporter.sendTelemetryEvent(event, data));
```

boaz-codota (approved):
Overall looks good, had some suggestions that I hope you fix before merge

src/reports/reporter.ts
```diff
@@ -11,24 +13,37 @@ export enum EventName {
...
   reporter = new TelemetryReporter(id, version, key);
   context.subscriptions.push(reporter);
 }
 
 export function report(event: EventName): void {
-  reporter.sendTelemetryEvent(event);
+  if (inTestMode) return;
+
+  void (async () => {
+    const data = await getReportData();
```
dimacodota, line 34: if any of the `systeminformation` calls  will be rejected, `sendTelemetryEvent` will not be called. is this the expected behavior?

src/reports/reporter.ts
```diff
@@ -11,24 +13,37 @@ export enum EventName {
...
+
   reporter = new TelemetryReporter(id, version, key);
   context.subscriptions.push(reporter);
 }
 
 export function report(event: EventName): void {
-  reporter.sendTelemetryEvent(event);
+  if (inTestMode) return;
+
+  void (async () => {
```
dimacodota, line 33: why the anonymous function ? why not  change the `report` function to be promise aware?

src/reports/reporter.ts
```diff
@@ -11,24 +13,37 @@ export enum EventName {
...
+    const data = await getReportData();
+    reporter.sendTelemetryEvent(event, data);
+  })();
 }
 
 export function reportErrorEvent(event: EventName, error: Error): void {
+  if (inTestMode) return;
+
   reporter.sendTelemetryErrorEvent(event, { error: error.message });
```
dimacodota, line 42: reportData should be added here as well

src/reports/reporter.ts
```diff
@@ -11,24 +13,37 @@ export enum EventName {
...
 }
 export function reportException(error: Error): void {
+  if (inTestMode) return;
+
   reporter.sendTelemetryException(error);
```
dimacodota, line 47: reportData should be added here as well

src/reports/reportData.ts
```diff
@@ -0,0 +1,85 @@
...
+  cpu: string;
+  cores: string;
+  speedGHz: string;
+  memoryGB: string;
+};
+
+let specsCache: Promise<Specs> | undefined;
+
+async function getSpecs(): Promise<Specs> {
+  const cpuData = await systeminformation.cpu();
```
dimacodota, line 42: use Promise.all()

src/reports/reportData.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  return specsCache;
+}
+
+export default async function getReportData(): Promise<ReportData> {
+  const specs = await getSpecsCache();
+
+  return {
+    timestamp: `${new Date().getTime()}`,
```
dimacodota, line 77: why there are 2 transformations( `osData` -> `Spec` -> `ReportData`)?

dimacodota (commented):
see my questions

src/reports/reporter.ts
yonip23: I think it makes the api cleaner this way, so you don't need to 'color' all caller functions / `void` it

src/reports/reporter.ts
yonip23: I think we should send the event without the properties, thanks

src/reports/reportData.ts
yonip23: `osData` is part of `Specs`, and `Specs` are not meant to be the final `ReportData` (more properties that are not spec related can be added potentially)

src/reports/reportData.ts
yonip23: why, its more readable this way

boaz-codota: it is not the same... `Promise.all` is "concurrent" (as any single request does not block the execution of that method, only once all requests happened then execution is blocked)

boaz-codota (approved):

src/reports/reportData.ts
yonip23: oh i see, makes sense. will change

dimacodota: by calling them this way(your, original ) you are actually  executing them in serial(the second call would not be called until first is resolved) it. is ok if  each call is dependent on the previous , now i'm not sure what `systeminformation` is doing under the hood and why the API is async, but if it is using some IO, then using Promise.all will handle these promises in parallel.

src/reports/reporter.ts
```diff
@@ -11,25 +13,44 @@ export enum EventName {
...
   reporter = new TelemetryReporter(id, version, key);
   context.subscriptions.push(reporter);
 }
 
 export function report(event: EventName): void {
-  reporter.sendTelemetryEvent(event);
+  if (inTestMode) return;
+
+  void getReportData().then((data) => {
+    reporter.sendTelemetryEvent(event, data ?? {});
```
dimacodota, line 34: why nullish ? sendTelemetryEvent arguments are optional

src/reports/reporter.ts
```diff
@@ -11,25 +13,44 @@ export enum EventName {
...
+      : { error: error.message };
+    reporter.sendTelemetryErrorEvent(event, fullData);
+  });
 }
 export function reportException(error: Error): void {
-  reporter.sendTelemetryException(error);
+  if (inTestMode) return;
+
+  void getReportData().then((data) => {
+    reporter.sendTelemetryException(error, data ?? {});
```
dimacodota, line 52: why nullish ? sendTelemetryEvent arguments are optional

src/reports/reporter.ts
```diff
@@ -11,25 +13,44 @@ export enum EventName {
...
+  void getReportData().then((data) => {
+    reporter.sendTelemetryEvent(event, data ?? {});
+  });
 }
 
 export function reportErrorEvent(event: EventName, error: Error): void {
-  reporter.sendTelemetryErrorEvent(event, { error: error.message });
+  if (inTestMode) return;
+
+  void getReportData().then((data) => {
```
dimacodota, line 41: set 'data'  to default  `{}` value and then just use `{ ...data, error: error.message }`

dimacodota (approved):
LGTM, had a very small suggestion

src/reports/reportData.ts
```diff
@@ -0,0 +1,98 @@
...
+export type ReportData = {
+  timestamp: string;
+  platform: string;
+  distro: string;
+  arch: string;
+  kernel: string;
+  cpu_manufacturer: string;
+  cpu_brand: string;
+  cores: string;
+  speedGHz: string;
```
bilucodota, line 38: `snake_case`

src/reports/reportData.ts
```diff
@@ -1,12 +1,20 @@
...
-import currentDateTime from "./utils";
+import { format, addMinutes } from "date-fns";
 
 const UNKNOWN_SPEED = -1;
 const UNKNOWN_CORES_AMOUNT = -1;
 const UNKNOWN_MEMORY_SIZE = -1;
 
 const byteToGigabyte = (bytes: number): number => bytes / 1e9;
 
+const currentDateTimeUTC: () => string = () => {
```
dimacodota, line 10: convert to regular function (not function expression)

dimacodota (approved):
lgtm, small suggestion

PR #448, opened 2021-04-01
handle published version validation

dimacodota:
handle published version validation

### DIFF
--- a/src/binary/binaryFetcher/bundleDownloader.ts
+++ b/src/binary/binaryFetcher/bundleDownloader.ts
@@ -1,6 +1,8 @@
 import { promises as fs } from "fs";
 import * as path from "path";
 import * as extract from "extract-zip";
+import * as semver from "semver";
+
 import {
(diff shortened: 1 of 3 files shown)


src/binary/binaryFetcher/bundleDownloader.ts
```diff
@@ -65,7 +67,15 @@ function createBundleDirectory(bundleDirectory: string): Promise<void> {
 
 async function getCurrentVersion(): Promise<string> {
   const versionUrl = getUpdateVersionFileUrl();
-  return downloadFileToStr(versionUrl);
+  const version = await downloadFileToStr(versionUrl);
+  validateVersion(version);
+  return version;
+}
+
+function validateVersion(version: string): void {
```
boaz-codota, line 75: rename to `assertValidVersion` so it will be easier to understand that this function throws an error

boaz-codota (approved):

PR #453, opened 2021-04-06
Add source to configuration requests

boaz-codota:


src/binary/requests/requests.ts
```diff
@@ -54,9 +54,10 @@ export function autocomplete(
   });
 }
 
-export function configuration(
-  body: { quiet?: boolean } = {}
-): Promise<{ message: string } | null | undefined> {
+export function configuration(body: {
+  quiet?: boolean;
+  source: string;
```
yonip23, line 59: Maybe create an enum for the `source`?

PR #468, opened 2021-04-08
Write binaries to globalStorageUrl & Add env variables to run

boaz-codota:


src/binary/paths.ts
```diff
@@ -1,35 +1,56 @@
...
 
+let binaryRootPath: string | undefined;
 const ARCHITECTURE = getArch();
 const SUFFIX = getSuffix();
 const BUNDLE_SUFFIX = getBundleSuffix();
 
+export function setBinaryRootPath(updatedPath: vscode.Uri): void {
+  binaryRootPath = path.join(updatedPath.fsPath, "binaries");
+  // eslint-disable-next-line no-debugger
+  debugger;
```
dimacodota, line 14: why?

src/binary/paths.ts
```diff
@@ -1,35 +1,56 @@
...
+  fs.mkdir(binaryRootPath, { recursive: true }, () => {});
```
dimacodota, line 15: mkdir is an async callback api, use `await fs.promises.mkdir`, if you want to be sure it is created

src/binary/paths.ts
```diff
@@ -1,35 +1,56 @@
...
+}
+
 export function versionPath(version: string): string {
-  return path.join(BINARY_ROOT_PATH, version, `${ARCHITECTURE}-${SUFFIX}`);
+  if (!binaryRootPath) {
```
dimacodota, line 19: extract to `assertBinaryRootPath`

dimacodota (changes requested):
see my comments

dimacodota (approved):
lgtm

Issue #325, opened 2021-02-05
Tabinine not working  : Activating extension 'TabNine.tabnine-vscode' failed: Cannot find module 'typemoq'


saurav9878:
- OS version: Ubuntu 20.04.1 LTS
- Editor version [e.g. 1.45.0-insider]: VSCode 1.53.0
- Programming language: [e.g. js, ts, c++] Javascript
- TabNine extension version: 3.2.6

Unable to open Tabnine::config

**Editor console logs:**
```
Activating extension 'TabNine.tabnine-vscode' failed: Cannot find module 'typemoq'
Require stack:
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/binary/mockedRunProcess.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/binary/runProcess.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/binary/runBinary.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/binary/Binary.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/binary/requests/requests.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/capabilities.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/alphaInstaller.js
- /home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/extension.js
- /usr/share/code/resources/app/out/vs/loader.js
- /usr/share/code/resources/app/out/bootstrap-amd.js
- /usr/share/code/resources/app/out/bootstrap-fork.js.
```
**Issue Details:**
 Initially, I thought the extension was not installed properly. I tried uninstalling, removing the folder inside extensions, and also local/share/tabnine models.
 Methods tried: I also went inside tabnine.tabnine-vscode-3.2.6 and ran npm install (tried even yarn) no luck. I was caught up with following error:
 ```
stack trace: TypeError: Cannot read property 'includes' of undefined
	at Object.executeStartupActions [as default] (/home/saurav/.vscode/extensions/tabnine.tabnine-vscode-3.2.6/out/binary/startupActionsHandler.js:10:93)
	at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

dimacodota:
[...]

PR #480, opened 2021-04-13
Fix tests and CI

boaz-codota:
Also migrated from npm to yarn (all of our other repos use yarn)


src/binary/paths.ts
```diff
@@ -11,7 +11,10 @@ const BUNDLE_SUFFIX = getBundleSuffix();
 export async function setBinaryRootPath(
   updatedPath: vscode.Uri
 ): Promise<void> {
-  binaryRootPath = path.join(updatedPath.fsPath, "binaries");
+  binaryRootPath =
+    process.env.NODE_ENV === "test"
```
dimacodota, line 15: lets extract it to some helper, it's used in several places

dimacodota: or even better, there is a `extensionMode` property on the `ExtensionContext` that we can use to determine if it is a test run. `NODE_ENV === test` can be modified locally, ref https://github.com/codota/tabnine-vscode/issues/325

.github/workflows/alpha_release.yml
```diff
@@ -28,16 +28,18 @@ jobs:
         base_version: "v${{ steps.package-version.outputs.current-version }}-alpha"
     - uses: actions/setup-node@v1
       with:
-        node-version: '10.x'
+        node-version: '12.x'
+    - name: Cache
+      uses: codota/gha-yarn-cache@v1
     - name: Install
-      run: npm ci --no-optional
+      run: yarn install --frozen-lockfile --ignore-optional
```
dimacodota, line 35: like

dimacodota (changes requested):
see my comments

dimacodota (approved):
lgtm

PR #539, opened 2021-06-20
Feature/beta channel

yonip23:


src/preRelease/versions.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function getAvailableAlphaVersion(artifactUrl: string): string {
+  const versionPattern = /(?<=download\/)(.*)(?=\/tabnine-vscode)/gi;
+  const match = artifactUrl.match(versionPattern);
+  return (match && match[0]) || "";
+}
+
+export function userConsumesAlphaVersions(): boolean {
+  const isVersionSupported = semver.gte(
```
dimacodota, line 37: you need to check this for beta as well

dimacodota (approved):
see my comment

src/globals/tabnineExtensionProperties.ts
```diff
@@ -59,6 +61,12 @@ function getContext(): TabNineExtensionProperties {
       true
     );
   }
+  const isExtentionBetaChannelEnabled =
+    configuration.get<boolean>("tabnine.receiveBetaChannelUpdates") || false;
+
+  const isVscodeInsiders =
+    vscode.env.appName.toLocaleLowerCase().indexOf("insider") > -1;
```
dimacodota, line 68: you can use .includes

src/preRelease/versions.ts
yonip23: Why is this needed? What isn't supported on previous versions?

PR #603, opened 2021-08-11
Task/tree view demo

dimacodota:


src/hub/hub.ts
```diff
@@ -70,6 +70,10 @@ export default function hub(): Hub {
               tempRef.parentNode.removeChild(tempRef);
               break;
             }
+            case "navigation": {
+              document.getElementById("config").src = data.url;
```
bilucodota, line 74: A bit more complicated - but you could have passed this into the iframe instead of replacing the url. Making the entire experience smoother.

src/hub/hubNavigation.ts
```diff
@@ -0,0 +1,17 @@
+import { env, Uri } from "vscode";
+import { configuration } from "../binary/requests/requests";
+import { StateType } from "../globals/consts";
+import openHub from "./openHub";
+
+export default async function open(view: string): Promise<void> {
```
bilucodota, line 6: `default exported function != file name`
generally speaking - the name open/hubNavigation is not clear

src/treeView/TabnineTreeProvider.ts
```diff
@@ -0,0 +1,29 @@
...
+  implements TreeDataProvider<TabnineTreeItem> {
+  onDidChangeTreeData?:
+    | Event<void | TabnineTreeItem | null | undefined>
+    | undefined;
+
+  getTreeItem(element: TabnineTreeItem): TreeItem {
+    return element as TreeItem;
+  }
+
+  getChildren(): ProviderResult<TabnineTreeItem[]> {
```
bilucodota, line 17: I understand this hook is called when user opens the view - maybe you can use it to open the hub on `home`?
Instead of letting the user click it

src/selectionHandler.ts
```diff
@@ -26,6 +26,8 @@ import { doPollNotifications } from "./notifications/pollNotifications";
 export const COMPLETION_IMPORTS = "tabnine-completion-imports";
 export const HANDLE_IMPORTS = "tabnine-handle-imports";
 
+// const shareDecorator = window.createTextEditorDecorationType({backgroundColor: "purple"});
```
bilucodota, line 29: remove?

PR #617, opened 2021-08-24
Task/for yoni

yonip23:


src/inlineSuggestions/setInlineSuggestion.ts
```diff
@@ -86,31 +91,58 @@ function clearPrefixFromSuggestion(currentCompletion: string, prefix: string) {
   return currentCompletion?.replace(prefix, "");
 }
 
-function showInlineDecoration(position: Position, suggestion: string): void {
-  const suggestionDecoration: DecorationOptions = {
+async function showInlineDecoration(
```
dimacodota, line 94: why not to split it into two - inlineDecoration and snippetDecoration?

yonip23: Because everything is the same except one extra step that's only relevant to snippets, so the flow makes more sense to me this way.

Also, there's a state that is managed between the creation of the decoration and the deletion (the `sniipetBlankRange` object or whatever i called it)

Wdyt?

PR #646, opened 2021-09-29
feature/short displayname

bilucodota:
- package.json
- swap to short display name when not in debug mode
- swap to short display name when not in debug mode


dimacodota (approved):
lgtm

PR #649, opened 2021-09-30
snippets metadata

yonip23:


src/inlineSuggestions/snippets/snippetProvider.ts
```diff
@@ -1,26 +1,58 @@
...
-    position,
-    "snippet",
-    trigger
-  );
+  const autocompleteResult:
+    | AutocompleteResultState
+    | undefined
+    | null = await runCompletion(document, position, "snippet", trigger);
+
+  if (autocompleteResult) autocompleteResult.onStateChange = onStateChange;
```
dimacodota, line 28: why the onStateChange is part of the autocompleteResult? it is not related to the autocompleteResult state

yonip23: its the `onStateChange` of this specific results batch, if for example the inline suggestions are setting the state with new results, i want the `onStateChange` to be removed

yonip23: if the state change is provided via a getter function, who would know when to call this function?
it should be called upon `alt+[ / ]`, creation of a new state and clearing of a state. i think that the best way to implement this is passing a callback that the inlineCompletionState module knows when to call, and the callback knows what to do.

PR #685, opened 2021-11-09
improve snippets behaviour

yonip23:
# Problems
1. Snippets are not deleted upon backspace
2. If you request a snippet in file a and then switch to file b - the snippet are appearing in the new location (file b). It happens because snippets are taking long to retrieve (1-2 sec).
Note: We do have a listener that clears the state upon cursor move, but its not triggered if you switch between files w/o moving the cursor.

# Solutions
1.  in `src/inlineSuggestions/registerHandlers.ts` - `e.kind === undefined` is true when the change is backspace, so not ignoring it and clearing the state in such cases. Its also true when the snippet blank space is being inserted, hence the `!isInSnippetInsertion()` check.
2. in `src/inlineSuggestions/snippets/snippetProvider.ts` - after retrieving snippet completions, check if the current file matches the file in which the request was fired.


dimacodota (approved):
lgtm

PR #689, opened 2021-11-10
feature: empty lines -> inline completions

yonip23:
### What's been done
If `EMPTY_LINE_SUGGESTIONS` capability is enabled, provide inline suggestions on empty lines

### Technical aspect
1. Move empty lines logic from `provideCompletionIterms.ts` (popup completions) to `textListener.ts` (inline completions)
2. Add `CompletionType` to `inlineSuggestionState.ts` to differentiate between snippet completions and inline completions - previously was done by checking if the completion contains more than one line, which is a weaker way to get this information.
3. extract `getCurrentPosition` to a new file, allowing both snippet and regular completions to use the same function to determine a `Position` from `TextDocumentContentChangeEvent`. While I was playing around with the logic of this function to support empty lines, it came out as just what was needed to unify snippets and regular completions so I just merged them.

### DIFF
--- a/src/inlineSuggestions/positionExtracter.ts
+++ b/src/inlineSuggestions/positionExtracter.ts
@@ -46,7 +46,7 @@ export function isOnlyWhitespaces(text: string): boolean {
 
 export function isEmptyLinesWithNewlineAutoInsert(
   change: TextDocumentContentChangeEvent
-) {
+): boolean {
   return getLines(change.text).length > 2 && isOnlyWhitespaces(change.text);
 }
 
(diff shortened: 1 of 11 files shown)


src/inlineSuggestions/positionExtracter.ts
```diff
@@ -0,0 +1,55 @@
+import { EOL } from "os";
```
dimacodota, line 1: `positionExtracter` ?

yonip23: do you have a better name? it extracts the position from the change...

dimacodota (approved):
lgtm

PR #694, opened 2021-11-16
snippets improvements

yonip23:
## Towards merging snippet completions with regular ones
We're trying to smoothly integrate snippets as part of your regular tabnine experience, which requires a lot of things (outside the scope of this pr).
### What's been done
- Use `CompletionKind` to determine whether the current inline hint is a snippet or a regular one - this allows to have a state with both types of completions and move between them smoothly
- Removed an old `if` statement in `src/inlineSuggestions/textListener.ts` that's turned out to be redundant:
The code was suppose to clear the inline hint in case the text has changed and you don't want to show a new hint. This is being handled by our cursor listener - a changed text also moves the cursor.
- Fix snippet hint deletion upon cursor move - in my previous fixes I caused a problem where if you move your cursor while a snippet hint is being displayed, it removes the same **range** that the snippet took, but **shifts** it to where your current cursor is.
For example, if you had:
```javascript
function foo() {
  // <- cursor here [snippet]
  // [snippet]
  // [snippet]
}
```
and you moved your cursor one line up, it'll delete the first line as well:
```javascript
// <- cursor here [deleted "function foo() {"]
// [deleted empty line]
// [deleted empty line]

}
```

### DIFF
--- a/src/inlineSuggestions/snippets/snippetDecoration.ts
+++ b/src/inlineSuggestions/snippets/snippetDecoration.ts
@@ -98,8 +98,9 @@ function calculateStartAfterUserInput(range: Range): Range | undefined {
   const textInsideSnippetBlankRange = window.activeTextEditor?.document.getText(
     range
   );
+  const blankRangeContainsText = textInsideSnippetBlankRange?.trim() !== "";
 
-  if (currentPosition && textInsideSnippetBlankRange?.trim() !== "") {
+  if (currentPosition && blankRangeContainsText) {
     const linesDiff = currentPosition.line - range.start.line;
     const charsDiff = currentPosition.character - range.start.character;
     return new Range(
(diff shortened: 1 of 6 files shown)


dimacodota (approved):
lgtm ❤️

src/inlineSuggestions/snippets/snippetDecoration.ts
```diff
@@ -95,8 +95,11 @@ export function handleClearSnippetDecoration(): void {
 
 function calculateStartAfterUserInput(range: Range): Range | undefined {
   const currentPosition = window.activeTextEditor?.selection.active;
+  const textInsideSnippetBlankRange = window.activeTextEditor?.document.getText(
+    range
+  );
 
-  if (currentPosition) {
+  if (currentPosition && textInsideSnippetBlankRange?.trim() !== "") {
```
dimacodota, line 102: `textInsideSnippetBlankRange?.trim() !== ""` is not so clear to me, maybe extract it to a named parameter

PR #716, opened 2021-12-07
fix: check minimal authentication provider api version

dimacodota:
check the minimal [`registerAuthenticationProvider`](https://code.visualstudio.com/updates/v1_54#_authentication-provider-api) version.  this api was supported in the  [1.54](https://code.visualstudio.com/updates/v1_54) version

### DIFF
--- a/src/globals/versions.ts
+++ b/src/globals/versions.ts
@@ -1,5 +1,6 @@
 import * as semver from "semver";
 import tabnineExtensionProperties from "./tabnineExtensionProperties";
+
 const AUTHENTICATION_API_VERSION = "1.54.0";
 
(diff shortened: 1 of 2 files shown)


src/globals/versions.ts
```diff
@@ -0,0 +1,10 @@
+import * as semver from "semver";
+import tabnineExtensionProperties from "./tabnineExtensionProperties";
+const AUTHENTICATION_API_VERSION = "1.54.0";
+
+export default function isAuthenticationApiSupported(): boolean {
```
yonip23, line 5: You can move the "minimal supported version" check we have in the pre release installer to here too

yonip23 (approved):
See my comment

PR #721, opened 2021-12-14
feature: debounce empty lines request

yonip23:
Note: need to implement it on the proposed api branch too. The file in which I added the debounce is deleted completely in the proposed api branch.
(Also need to to take into consideration the fact the snippets are taking longer to fetch to begin with...)

### DIFF
--- a/src/inlineSuggestions/textListener.ts
+++ b/src/inlineSuggestions/textListener.ts
@@ -27,9 +27,14 @@ const EMPTY_LINE_WARMUP_MILLIS = 250;
 const EMPTY_LINE_DEBOUNCE = 550;
 
 const debouncedEmptyLinesRequest = debounce(
-  async (document: TextDocument, currentPosition: Position) => {
-    const autocompleteResult = await runCompletion(document, currentPosition);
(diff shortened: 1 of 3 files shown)


src/inlineSuggestions/textListener.ts
```diff
@@ -18,8 +21,18 @@ import getCurrentPosition, {
   isEmptyLinesWithNewlineAutoInsert,
   isOnlyWhitespaces,
 } from "./positionExtracter";
+import { AutocompleteResult } from "../binary/requests/requests";
 
-const EMPTY_LINE_WARMUP_MILLIS = 110;
+const EMPTY_LINE_WARMUP_MILLIS = 250;
+const EMPTY_LINE_DEBOUNCE = 550;
+
+const debouncedEmptyLinesRequest = debounce(
```
dimacodota, line 29: debounce is not promise aware, https://github.com/component/debounce/blob/8be734c4b5b5a58a517f2441e0d90242a4a564cd/index.js#L27
meaning you cant pass an async function(at least not without handling rejections)

dimacodota (approved):
lgtm

PR #735, opened 2021-12-29
adjust gitpod persistent state solution to codespaces

yairco1990:


src/globals/consts.ts
```diff
@@ -131,7 +131,7 @@ export function restartBackoff(attempt: number): number {
 
 export const IS_OSX = process.platform === "darwin";
 
-export const SLEEP_TIME_BEFORE_OPEN_HUB = isGitpod ? 1000 * 10 : 0;
+export const SLEEP_TIME_BEFORE_OPEN_HUB = isCloudEnv ? 1000 * 10 : 0;
```
bilucodota, line 134: Not sure we need to wait here in case of codespaces. But this okay for 1st phase.

bilucodota (approved):

PR #737, opened 2022-01-03
feat: use graceful item description instead of large window

zardoy:
I think there should be a strict rule in vscode extensions guidelines: *every extension that contributes suggestions should set description on each item*. Obviously this makes items distinguishable between different extensions. However you guys already had BRAND_ICON in label start. So the main intention of this change is to get rid of displaying detail window. Detail windows should be used only for displaying additional code-related info (though I know that even builtin TS extension doesn't follow this rule).
This is how it looks like (notice, side-window isn't displayed anymore):
image

As I remeber you can set label description only via constructor and this wasn't available in `1.50.0` so I updates the types. I didn't resolve all types errors, there are still in `TabnineAuthenticationProvider.ts`, hope to see some help here. I can also revert the first commit, and wait until you update vscode types.

### DIFF
--- a/src/provideCompletionItems.ts
+++ b/src/provideCompletionItems.ts
@@ -83,14 +83,12 @@ function makeCompletionItem(args: {
   results: ResultEntry[];
   limited: boolean;
 }): vscode.CompletionItem {
-  const item = new vscode.CompletionItem(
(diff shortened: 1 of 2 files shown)


zardoy:
@dimacodota

PR #745, opened 2022-01-12
Add notifications widget

ofekby:


src/manageTeam/manageTeamWebview.ts
```diff
@@ -0,0 +1,66 @@
...
+  const provider = new ManageTeamWebviewProvider();
+  void setManageTeamWebviewReady();
+
+  context.subscriptions.push(
+    window.registerWebviewViewProvider("tabnine-team", provider)
+  );
+}
+
+async function setManageTeamWebviewReady() {
+  if (await isNewerThanVersion([4, 0, 55])) {
```
dimacodota, line 23: use [isCapabilityEnabled](https://github.com/codota/tabnine-vscode/blob/5840dee427f0f369f51a1a249d843bd6a0d70203/src/capabilities/capabilities.ts#L26) instead, it is controlled via the https://github.com/codota/acpl-paid/blob/master/features.toml file

dimacodota (approved):
lgtm

src/treeView/registerTreeView.ts
```diff
@@ -1,31 +1,38 @@
...
+        new TabnineTreeProvider()
+      ),
+      commands.registerCommand(TABNINE_TREE_NAVIGATION_COMMAND, (view) => {
+        void navigate(view);
+        void fireEvent({
+          name: "hub-opened-from-sidebar",
+        });
+      }),
+      commands.registerCommand(TABNINE_OPEN_APP_COMMAND, () => {
+        void env.openExternal(Uri.parse(TABNINE_APP_URL));
```
dimacodota, line 25: maybe it would be better to route it through the binary ? by default vscode will inform the user that it is about to open a url (via a modal popup)

src/notificationsWidget/manageTeamWebview.ts
```diff
@@ -0,0 +1,73 @@
...
+  window,
+} from "vscode";
+import layout from "../utils/webviewLayout";
+import { getHubBaseUrl } from "../utils/binary.utils";
+import { Capability, isCapabilityEnabled } from "../capabilities/capabilities";
+import { fireEvent } from "../binary/requests/requests";
+
+function registerNotificaitonsWebviewProvider(context: ExtensionContext): void {
+  const provider = new NotificationsWebviewProvider();
+  void setManageTeamWebviewReady();
```
dimacodota, line 15: why not after the provider is registered (registerWebviewViewProvider)?

src/notificationsWidget/manageTeamWebview.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  context.subscriptions.push(
+    window.registerWebviewViewProvider("tabnine-notifications", provider)
+  );
+}
+
+function setManageTeamWebviewReady() {
+  if (isCapabilityEnabled(Capability.NOTIFICATIONS_WIDGET)) {
```
dimacodota, line 23: isCapabilityEnabled can throw an exception, and u r void(ing) on `setManageTeamWebviewReady` call

src/treeView/registerTreeView.ts
dimacodota: ping

src/notificationsWidget/manageTeamWebview.ts
```diff
@@ -0,0 +1,73 @@
...
+class NotificationsWebviewProvider implements WebviewViewProvider {
+  // eslint-disable-next-line class-methods-use-this
+  resolveWebviewView(webviewView: WebviewView): void | Thenable<void> {
+    // eslint-disable-next-line no-param-reassign
+    webviewView.webview.options = {
+      enableScripts: true,
+      enableCommandUris: true,
+    };
+
+    return (async () => {
```
dimacodota, line 41: extract to function

src/notificationsWidget/manageTeamWebview.ts
ofekby: It cant throw exception - but I changed the order and remove the void

PR #747, opened 2022-01-15
add snippet shown event

yonip23:
### Update

The below issue has been solved by using the new api version 4.0.57

---

@dimacodota same as here: https://github.com/codota/tabnine-intellij/pull/263#issue-1101626578

### DIFF
--- a/src/inlineSuggestions/registerHandlers.ts
+++ b/src/inlineSuggestions/registerHandlers.ts
@@ -11,6 +11,7 @@ import {
   window,
   workspace,
 } from "vscode";
+import setState from "../binary/requests/setState";
 import { Capability, isCapabilityEnabled } from "../capabilities/capabilities";
 import getSuggestionMode, {
   SuggestionsMode,
@@ -21,6 +22,7 @@ import {
   NEXT_INLINE_COMMAND,
   PREV_INLINE_COMMAND,
   SNIPPET_COMMAND,
+  StatePayload,
 } from "../globals/consts";
 import enableProposed from "../globals/proposedAPI";
 import provideInlineCompletionItems from "../provideInlineCompletionItems";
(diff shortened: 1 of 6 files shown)


src/inlineSuggestions/registerHandlers.ts
```diff
@@ -88,6 +91,14 @@ export default async function registerInlineHandlers(
...
       console.log("onDidChangeTextEditorSelection: ", e.kind);
     });
+    window
+      .getInlineCompletionItemController(inlineCompletionsProvider)
+      .onDidShowCompletionItem((e) => {
+        const shownCompletionIsMultiline = e.completionItem.text.includes("\n");
+        if (shownCompletionIsMultiline) {
+          void setState({ [StatePayload.SNIPPET_SHOWN]: {} });
+        }
+      });
```
yonip23, line 101: @dimacodota idk if its cool to use this api:
image

dimacodota: why not use the  `provideInlineCompletionItems`?

yonip23: because there we're only returning the results, there's an entire event loop iteration that happens between that and when the plugin actually shows the hint.. im guessing that if you move the cursor and such it could send false positives

src/inlineSuggestions/registerHandlers.ts
```diff
@@ -94,8 +95,11 @@ export default async function registerInlineHandlers(
     window
       .getInlineCompletionItemController(inlineCompletionsProvider)
       .onDidShowCompletionItem((e) => {
-        const shownCompletionIsMultiline = e.completionItem.text.includes("\n");
-        if (shownCompletionIsMultiline) {
+        const shouldSendSnippetShown =
+          e.completionItem.completionKind === CompletionKind.Snippet &&
+          !e.completionItem.isCached;
```
yonip23, line 100: this is the usage of `is_cached`

dimacodota (approved):
lgtm

PR #753, opened 2022-01-23
Yonip/add hint color config

yonip23:
image

### DIFF
--- a/src/inlineSuggestions/inlineSuggestionState.ts
+++ b/src/inlineSuggestions/inlineSuggestionState.ts
@@ -1,4 +1,4 @@
-import { commands } from "vscode";
+import { commands, workspace } from "vscode";
 import { AutocompleteResult, ResultEntry } from "../binary/requests/requests";
 import { rotate } from "../utils/rotate";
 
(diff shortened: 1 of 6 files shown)


yonip23:
@dimacodota a non-valid color in this field will result in a default color, which is white:
image

image

src/inlineSuggestions/inlineSuggestionState.ts
```diff
@@ -55,3 +55,10 @@ export function getAllSuggestions(): ResultEntry[] | undefined {
 function results(): ResultEntry[] | undefined {
   return autocompleteResult?.results;
 }
+
+export function getHintColor(): string {
```
dimacodota, line 59: why is it here ?

dimacodota (approved):
see my comment

src/inlineSuggestions/inlineSuggestionState.ts
yonip23: thought it can be considered as part of the suggestions state - the color's state which is being pulled from the settings. Should i move this to a new file?

PR #780, opened 2022-02-16
repackaging feb22 - show 'tabnine team' in statusbar

talcodota:
need to show "tabnine team" for paid customers, but show "tabnine pro" for legacy pro + students


boaz-codota (approved):
LGTM

PR #797, opened 2022-03-21
use api version 4.1.21

yonip23:
- add filename and intent to `SnippetShown`
- add intent to `AutocompleteResult` and to `SelectionRequest`


dimacodota (approved):
lgtm

PR #807, opened 2022-03-31
fix negative length in completion event data

yonip23:


src/selectionHandler.ts
```diff
@@ -135,16 +140,18 @@ function eventDataOf(
   });
 
   const { length } = currentCompletion;
-  const netLength = editor.selection.anchor.character - position.character;
+  const netLength = length - (oldPrefix?.length || 0);
```
yonip23, line 143: this is the main change - `netLength` was calculated with positions, which is problematic if the completion takes more than one line. i changed it to be based on the completion parameters themselves (`old_prefix` is quite exactly the opposite of what we mean by `netLength`)

src/inlineSuggestions/inlineSuggestionState.ts
```diff
@@ -52,6 +52,13 @@ export function getAllSuggestions(): ResultEntry[] | undefined {
   return results();
 }
 
+export function currentAutocompleteResult():
```
yonip23, line 55: remove

src/selectionHandler.ts
yonip23: this is also how we calculate it in the intellij plugin: https://github.com/codota/tabnine-intellij/blob/5c66c60f08d60463cf87b372264b3638c6faee1a/src/main/java/com/tabnine/selections/CompletionPreviewListener.java#L27

dimacodota (approved):
lgtm

dimacodota:
add some tests

PR #816, opened 2022-04-20
add debug context

dimacodota:
debug context added for "inline" suggestion, will be triggered for alpha capability only

Screen Shot 2022-04-20 at 14 45 19


yonip23 (commented):
wouldn't it work only for the old api? we're all working with the proposed api internally

dimacodota:
> wouldn't it work only for the old api? we're all working with the proposed api internally

yes, it is only relevant for the "old" api, the proposed api is only enabled if the "advanced completions" capability is enabled

yonip23 (approved):

PR #817, opened 2022-04-20
Feature/code review

amircodota:


src/codeReview/suggestions.ts
```diff
@@ -0,0 +1,137 @@
...
+async function supportedExtensions(): Promise<string[]> {
+  if (!cachedSupportedExtensions) {
+    cachedSupportedExtensions = (await api.supportedExtensions()).extensions;
+  }
+
+  return cachedSupportedExtensions;
+}
+
+setInterval(() => {
+  cachedSupportedExtensions = null;
```
dimacodota, line 136: maybe we can use axios-cache-adapter or something?

src/extension.ts
```diff
@@ -66,6 +67,8 @@ export async function activate(
   handleSelection(context);
   handleUninstall(() => uponUninstall(context));
 
+  registerCodeReview();
```
dimacodota, line 70: this is too soon(you wont get the capabilities here ), the capabilities are fetched in the `backgroundInit`   https://github.com/codota/tabnine-vscode/blob/3b9e6fa06148544a4ef226b99a63dc64cbd7e57b/src/extension.ts#L100

src/codeReview/suggestions.ts
```diff
@@ -0,0 +1,137 @@
...
+      lineCount: document.lineCount,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+      error: e.message || e.toString(),
+    });
+
+    throw e;
+  }
+}
+
+export class DocumentThreads {
```
dimacodota, line 111: move to file

package.json
```diff
@@ -251,8 +271,30 @@
         {
           "command": "TabNine::assistantToggle",
           "when": "tabnine-assistant:capability"
-        }
-      ]
+        },
+        {
+					"command": "Tabnine.hideSuggestion",
```
dimacodota, line 276: why do you need this declaration here?

amircodota: so it does not appear in the command pallete?

Saw that's what they did in the comments api sample

src/codeReview/TabnineComment.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    return true;
+  }
+
+  hide(thread: CommentThread): void {
+    this.fireEvent("comment-hidden", thread);
+    thread.dispose();
+  }
+
+  private static documentOf(thread: CommentThread): TextDocument | undefined {
```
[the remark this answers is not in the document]
amircodota: because it does not use this. The linter is not passing if it is not static

src/extension.ts
amircodota: I'm checking the capabilities only on activeEditorChanged event

package.json
dimacodota: i think you can remove it from here

src/codeReview/TabnineComment.ts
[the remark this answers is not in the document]
dimacodota: yea, it just seems like a helper standalone function witch is not related to the class

dimacodota (approved):
lgtm

src/codeReview/suggestions.ts
amircodota: I prefer it this way, because the api is a clean wrapper.

The one using it knows it can afford some staleness, so it caches the results, but if someone else uses the api, it might not want it cached

PR #828, opened 2022-05-03
trim snippets with respect to indentation

yonip23:
See the description on pr 1805 in our binary project for more details.

### DIFF
--- a/src/binary/requests/completionPostProcess.ts
+++ b/src/binary/requests/completionPostProcess.ts
@@ -1,0 +1,70 @@
+import {
+  AutocompleteParams,
+  AutocompleteResult,
+  CompletionKind,
+} from "./requests";
+
+export default function postprocess(
+  request: AutocompleteParams,
+  result: AutocompleteResult,
+  tabSize: number
+) {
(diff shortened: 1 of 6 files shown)


src/binary/requests/completionPostProcess.ts
```diff
@@ -0,0 +1,79 @@
...
+): void {
+  const resultsSubset = result.results.filter(
+    (entry) => entry.completion_kind === CompletionKind.Snippet
+  );
+  if (resultsSubset.length === 0) return;
+
+  resultsSubset.forEach((entry, index) => {
+    resultsSubset[index].new_prefix = entry.new_prefix.replace(
+      /\t/g,
+      " ".repeat(tabSize)
```
assapir, line 20: What if using tabs and not spaces for ident?

yonip23: We're always returning indentation in spaces in multiline suggestions, assuming the ides know how to handle it - im just normalizing the suggestion to this standard, because we're working now with some server that return tabs from time to time

assapir (approved):

PR #834, opened 2022-05-15
remove the beta check and add an 'advanced completion' check

dimacodota:
remove the beta check and add an 'advanced completion' check


yonip23 (approved):
lgtm

PR #838, opened 2022-05-18
Bug fix: Vim for VSCode mode switch on accept completion

LeshemNoa:
https://tabnine.atlassian.net/browse/DEV2-565

I played with it, tried new line, mid line and end of line completions, and it seems to work well. Regarding tests - I'm looking into it, not entirely sure how to install the vim extension in the test environment.

### DIFF
--- a/src/inlineSuggestions/acceptInlineSuggestion.ts
+++ b/src/inlineSuggestions/acceptInlineSuggestion.ts
@@ -1,4 +1,11 @@
-import { commands, Position, Range, SnippetString, TextEditor } from "vscode";
+import {
+  commands,
+  Position,
+  Range,
+  SnippetString,
+  TextEditor,
+  extensions,
+} from "vscode";
 import { ResultEntry } from "../binary/requests/requests";
 import { CompletionArguments } from "../CompletionArguments";
 import { COMPLETION_IMPORTS } from "../selectionHandler";
@@ -36,6 +43,26 @@ export default async function acceptInlineSuggestion(
           editor,
           prefix
         ));
+
+    const vimActive = extensions.getExtension("vscodevim.vim")?.isActive;
+    if (vimActive) {
+      await vimReturnToInsertMode(currentSuggestion);
+    }
+  }
+}
+
+async function vimReturnToInsertMode(suggestion: ResultEntry) {
+  await commands.executeCommand("extension.vim_escape");
+  await commands.executeCommand("extension.vim_insert");
+  const suggestionString = suggestion.new_prefix + suggestion.new_suffix;
+  vimMoveCursorRight(suggestionString.length);
+}
+
+function vimMoveCursorRight(steps: number) {
+  let count = 0;
+  while (count < steps) {
+    void commands.executeCommand("extension.vim_right");
+    count += 1;
   }
 }
 
--- a/src/inlineSuggestions/acceptInlineSuggestion.ts
+++ b/src/inlineSuggestions/acceptInlineSuggestion.ts
@@ -1,11 +1,4 @@
-import {
-  commands,
-  Position,
-  Range,
-  SnippetString,
-  TextEditor,
-  extensions,
-} from "vscode";
+import { commands, Position, Range, SnippetString, TextEditor } from "vscode";
 import { ResultEntry } from "../binary/requests/requests";
 import { CompletionArguments } from "../CompletionArguments";
 import { COMPLETION_IMPORTS } from "../selectionHandler";
@@ -17,6 +10,7 @@ import {
   getAllSuggestions,
 } from "./inlineSuggestionState";
 import acceptSnippetSuggestion from "./snippets/acceptSnippetSuggestion";
+import { vimActive, vimReturnToInsertMode } from "./vimForVSCodeWorkaround";
 
 export default async function acceptInlineSuggestion(
   editor: TextEditor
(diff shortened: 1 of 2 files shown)


src/inlineSuggestions/acceptInlineSuggestion.ts
```diff
@@ -36,9 +43,28 @@ export default async function acceptInlineSuggestion(
...
   }
 }
 
+async function vimReturnToInsertMode(suggestion: ResultEntry) {
+  await commands.executeCommand("extension.vim_escape");
+  await commands.executeCommand("extension.vim_insert");
+  const nextCursorPosition =
+    suggestion.new_prefix.length + suggestion.new_suffix.length;
+  setImmediate(() => {
+    void commands.executeCommand("cursorMove", {
```
dimacodota, line 60: extract to `moveCursorToEnd`

src/inlineSuggestions/acceptInlineSuggestion.ts
```diff
@@ -36,9 +43,28 @@ export default async function acceptInlineSuggestion(
...
+    }
   }
 }
 
+async function vimReturnToInsertMode(suggestion: ResultEntry) {
+  await commands.executeCommand("extension.vim_escape");
+  await commands.executeCommand("extension.vim_insert");
+  const nextCursorPosition =
+    suggestion.new_prefix.length + suggestion.new_suffix.length;
+  setImmediate(() => {
```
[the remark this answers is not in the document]
LeshemNoa: What happens without this is that the cursor jumps to the start of the completion after this callback is finished. Something moves the cursor again before the tick ends, and as far as I can tell it's not Tabnine because the cursor movement listener isn't triggered when it happens. Moving the cursor on the next tick seems to address that and work well.

dimacodota:
let's see if we can test it

src/inlineSuggestions/acceptInlineSuggestion.ts
```diff
@@ -36,9 +43,32 @@ export default async function acceptInlineSuggestion(
           editor,
           prefix
         ));
+
+    const vimActive = extensions.getExtension("vscodevim.vim")?.isActive;
```
yonip23, line 47: can this be a global const?

src/inlineSuggestions/acceptInlineSuggestion.ts
```diff
@@ -36,9 +43,32 @@ export default async function acceptInlineSuggestion(
...
+  await commands.executeCommand("extension.vim_escape");
+  await commands.executeCommand("extension.vim_insert");
+  const nextCursorPosition =
+    suggestion.new_prefix.length + suggestion.new_suffix.length;
+  setImmediate(() => {
+    moveCursorToPosition(nextCursorPosition);
+  });
+}
+
+function moveCursorToPosition(nextCursorPosition: number) {
```
yonip23, line 64: a better name imo can be `moveCursorRight(steps: number)`
or something that conveys the fact that this function moves the cursor to a relative position to the right

src/inlineSuggestions/acceptInlineSuggestion.ts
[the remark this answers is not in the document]
LeshemNoa: Turns out that we need to use the `vim` move cursor commands to achieve the desired result without manipulating the event order further. Hurray!

src/inlineSuggestions/acceptInlineSuggestion.ts
yonip23: @LeshemNoa

src/inlineSuggestions/acceptInlineSuggestion.ts
```diff
@@ -36,6 +43,26 @@ export default async function acceptInlineSuggestion(
...
         ));
+
+    const vimActive = extensions.getExtension("vscodevim.vim")?.isActive;
+    if (vimActive) {
+      await vimReturnToInsertMode(currentSuggestion);
+    }
+  }
+}
+
+async function vimReturnToInsertMode(suggestion: ResultEntry) {
```
yonip23, line 54: use destruction:
```suggestion
async function vimReturnToInsertMode({ new_prefix, new_suffix }: ResultEntry) {
```

yonip23 (approved):
i'd also extract all vim related logic to a separate file.
great job! 👏🏻

src/inlineSuggestions/acceptInlineSuggestion.ts
LeshemNoa: Fixed. PTAL

src/inlineSuggestions/vimForVSCodeWorkaround.ts
```diff
@@ -0,0 +1,25 @@
+import { commands, extensions } from "vscode";
+import { ResultEntry } from "../binary/requests/requests";
+
+export const VIM_FOR_VSCODE_EXTENSION = "vscodevim.vim";
```
yonip23, line 4: redundant `export` i think

src/inlineSuggestions/vimForVSCodeWorkaround.ts
```diff
@@ -0,0 +1,25 @@
...
+export function vimActive(): boolean {
+  return !!extensions.getExtension(VIM_FOR_VSCODE_EXTENSION)?.isActive;
```
yonip23, line 6: redundant `!!` i think

src/inlineSuggestions/vimForVSCodeWorkaround.ts
```diff
@@ -0,0 +1,25 @@
...
+}: ResultEntry): Promise<void> {
+  await commands.executeCommand("extension.vim_escape");
+  await commands.executeCommand("extension.vim_insert");
+  const suggestionString = new_prefix + new_suffix;
+  vimMoveCursorRight(suggestionString.length);
+}
+
+function vimMoveCursorRight(steps: number): void {
+  let count = 0;
+  while (count < steps) {
```
yonip23, line 21: can be a `for` loop

PR #841, opened 2022-05-24
Feature/enable completions UX change in realtime

amircodota:
Support having the inline capabilities change affect immediately, without requiring a restart

### DIFF
--- a/src/capabilities/capabilities.ts
+++ b/src/capabilities/capabilities.ts
@@ -71,8 +71,8 @@ async function refreshCapabilities(): Promise<void> {
 let interval: NodeJS.Timeout | null = null;
 
 function startRefreshLoop(): void {
-  let lastPid = tabNineProcess.pid(),
-    lastRefresh = new Date();
+  let lastPid = tabNineProcess.pid();
+    let lastRefresh = new Date();
 
   if (interval) {
     clearInterval(interval);
--- a/src/capabilities/capabilities.ts
+++ b/src/capabilities/capabilities.ts
@@ -1,5 +1,9 @@
 import * as vscode from "vscode";
-import { getCapabilities } from "../binary/requests/requests";
+import { getCapabilities, tabNineProcess } from "../binary/requests/requests";
+import { getTabnineExtensionContext } from "../globals/tabnineExtensionContext";
+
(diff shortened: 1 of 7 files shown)


src/capabilities/capabilities.ts
```diff
@@ -49,11 +54,52 @@ export function fetchCapabilitiesOnFocus(): Promise<void> {
...
+
+export function onDidRefreshCapabilities(listener: () => void): Disposable {
+  return capabilitiesRefreshed.event(listener);
+}
+
+async function refreshCapabilities(): Promise<void> {
+  const capabilities = await getCapabilities();
+
+  enabledCapabilities = {};
+  capabilities?.enabled_features.forEach((feature) => {
```
dimacodota, line 73:
```suggestion
  enabledCapabilities = capabilities?.enabled_features.map((feature) => ({[feature]: true})
```

amircodota: This does not do the same thing

```
> let enabled_features = ['a', 'b', 'c'];
undefined
> 
> let enabledCapabilities = {};
undefined
> enabled_features.forEach((feature) => {
...     enabledCapabilities[feature] = true;
... });
undefined
> 
> enabledCapabilities
{ a: true, b: true, c: true }
> 
> enabled_features.map((feature) => ({[feature]: true}))
[ { a: true }, { b: true }, { c: true } ]
```

To use it properly, we can do something like

```
Object.assign({}, ...enabled_features.map((feature) => ({[feature]: true})))
```

Or maybe use `Array.reduce`

```
enabled_features.reduce((o, feature) => {
	o[feature] = true;
	return o;
}, {});
```

Both do not seem more readable to me than the original

src/capabilities/capabilities.ts
```diff
@@ -49,11 +54,52 @@ export function fetchCapabilitiesOnFocus(): Promise<void> {
...
-
+  void refreshCapabilities().then(() => {
     resolve();
+    startRefreshLoop();
+  });
+}
+
+const capabilitiesRefreshed = new EventEmitter<void>();
+
+export function onDidRefreshCapabilities(listener: () => void): Disposable {
```
dimacodota, line 65: it is bit strange that the `listener` does not get the refreshed capabilities as input, isn't it?

dimacodota (approved):
lgtm

PR #855, opened 2022-06-09
refresh capabilities more often

yonip23:
in order to catch binary config changes in somewhat "real time", we shorten the interval between capabilities refresh


dimacodota (approved):
lgtm

PR #861, opened 2022-06-14
remove the unstable onDidShowCompletionItem api + use the new inline …

dimacodota:
remove the unstable onDidShowCompletionItem api + use the new inline if supported


amircodota (approved):
LGTM

PR #869, opened 2022-06-23
Open hub welcome page inside vscode

ofekby:


src/extension.ts
```diff
@@ -144,6 +152,20 @@ async function backgroundInit(context: vscode.ExtensionContext) {
...
+  if (!(await touchIsJustInstalledFlag(context))) {
+    return false;
+  }
+
+  const binaryState = await getBinaryState();
+
+  return (
+    binaryState?.installationTime &&
+    binaryState.flags?.shouldUseNewHubDesign &&
+    isInTheLastHour(new Date(binaryState?.installationTime))
```
yonip23, line 165: i think you can use `binaryState.installationTime` without `?` here (because of the check in line 163)

src/installationFlag.ts
```diff
@@ -0,0 +1,17 @@
+import { ExtensionContext } from "./preRelease/types";
+
+const IS_JUST_INSTALLED_FLAG = "is-just-installed-tabnine";
```
yonip23, line 3: flag name is confusing - you set this key to `true` once you already handled the "just installed" logic, so it represent the opposite of `is-just-installed-tabnine` - its more like `already-installed` or something similar

src/utils/getBinaryState.ts
```diff
@@ -0,0 +1,26 @@
...
+  };
+}
+
+export default async function getBinaryState(): Promise<
+  PartialStateResponse | undefined
+> {
+  const binaryHttpServerUrl = await getHubBaseUrl(StateType.STARTUP);
+
+  if (!binaryHttpServerUrl) {
+    return undefined;
```
yonip23, line 18: i think returning `null` is more meaningful here

src/utils/time.utils.ts
```diff
@@ -0,0 +1,7 @@
+const MS_IN_SECOND = 1000;
+const MS_IN_MINUTE = 60 * MS_IN_SECOND;
+const MS_IN_HOUR = 60 * MS_IN_MINUTE;
```
yonip23, line 3:
```suggestion
const SECOND_IN_MS = 1000;
const MINUTE_IN_MS = 60 * MS_IN_SECOND;
const HOUR_IN_MS = 60 * MS_IN_MINUTE;
```

src/utils/getBinaryState.ts
```diff
@@ -0,0 +1,25 @@
...
+interface PartialStateResponse {
+  installationTime?: string;
+  enabledFeatures: string[];
+  flags?: {
+    shouldUseNewHubDesign?: boolean;
+  };
+}
+
+export default async function getBinaryState(): Promise<PartialStateResponse | null> {
+  const binaryHttpServerUrl = await getHubBaseUrl(StateType.STARTUP);
```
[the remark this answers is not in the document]
dimacodota: ie https://github.com/codota/tabnine-vscode/blob/297b641886ab0ecb3d35607755696bdf1e4c9065/src/binary/requests/requests.ts#L105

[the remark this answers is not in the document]
ofekby: Wanted to avoid adding the enabledFeatures to the State command - but I will change it now because I already used it now...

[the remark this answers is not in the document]
dimacodota: oh, but the `enabledFeatures` are returned as part of the `Capabilities` request

PR #872, opened 2022-06-27
fix hover position

dimacodota:
this PR fixes two issues :
 -the hint was displayed in the wrong position if the suggestion was multi-line
 -the hover tooltip "shown" event was triggered on any element and not on the relevant hint

### DIFF
--- a/src/hovers/decorationState.ts
+++ b/src/hovers/decorationState.ts
@@ -20,7 +20,7 @@ let decoration: DecorationOptions | null | undefined;
 let decorationsDebounce: NodeJS.Timeout | null | undefined;
 let currentHover: Hover | null | undefined;
 
-export function getDecoration(): Hover | null | undefined {
+export function getCurrentHover(): Hover | null | undefined {
   return currentHover;
 }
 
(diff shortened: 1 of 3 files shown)


src/hovers/decorationState.ts
```diff
@@ -18,12 +18,18 @@ const decorationType = window.createTextEditorDecorationType({
 
 let decoration: DecorationOptions | null | undefined;
 let decorationsDebounce: NodeJS.Timeout | null | undefined;
+let currentHover: Hover | null | undefined;
+
+export function getDecoration(): Hover | null | undefined {
```
yonip23, line 23: rename to `getCurrentHover` maybe?

src/selectionHandler.ts
```diff
@@ -86,7 +86,7 @@ export function getSelectionHandler(
       void setState(eventData).then(() => {
         void doPollNotifications(context);
         void doPollStatus(context);
-        void setHover(context, position);
+        void setHover(context, editor.selection.active.translate(0,10));
```
yonip23, line 89: extract `editor.selection.active.translate(0,10)` to a function, something that'll look like `marginRight(position)`

yonip23 (approved):

PR #877, opened 2022-07-04
Add file extension to untitled files' filenames using VSCode autodetection

LeshemNoa:
Piggybacking VSCode's ML based language autodetection on untitled files, we can provide a better experience on such files.


src/runCompletion.ts
```diff
@@ -52,3 +53,22 @@ function getMaxResults(): number {
...
+
+export function getFileNameWithExtension(
+  document: TextDocument
+): string {
+  const { languageId } = document;
+  let filenameWithExtension = document.fileName;
+  if (document.isUntitled) {
+    if (isKnownLanguage(languageId)) {
+      filenameWithExtension = document.fileName.concat(languages[languageId]);
+    }
```
yonip23, line 71: i'd create a function `getFileExtension(languageId: string): string | undefined` instead of `isKnownLanguage`, so that this code can look like so:

```typescript
const fileExtension = getFileExtension(languageId);
if (fileExtension) {
    filenameWithExtension = document.fileName.concat(fileExtension);
}
```

LeshemNoa:
@dimacodota the file was constructed manually, combining data from
- [the VSCode language identifier list](https://code.visualstudio.com/docs/languages/identifiers) 
- The output returned from `languages.getLanguages`, which surprisingly is not 1-1 with the ID list they documented themselves. Documented [here](https://code.visualstudio.com/api/references/vscode-api#languages)
- Our own `languages.yaml` from acpl.

LeshemNoa:
@yonip23 I don't understand what I did to cause this test case to fail, it seems totally unrelated! 🤡 ʜᴇʟᴘ

src/test/suite/untitledFile.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const res = await activate();
+    editor = res?.editor;
+    doc = res?.doc;
+  });
+
+  afterEach(async () => {
+    await vscode.commands.executeCommand("workbench.action.closeActiveEditor");
+  });
+
+  const tableIt = async (input: string, fileExtension: string) => {
```
yonip23, line 22: change to a meaningful name, something like `testUndefinedFileWith`

src/test/suite/utils/helper.ts
```diff
@@ -14,10 +14,12 @@ export type BinaryGenericRequest<T> = {
...
  */
 export async function activate(
-  docUri: vscode.Uri
+  docUri?: vscode.Uri
 ): Promise<{ editor: TextEditor; doc: TextDocument } | null> {
   try {
-    const doc = await vscode.workspace.openTextDocument(docUri);
+    const doc = docUri
+      ? await vscode.workspace.openTextDocument(docUri)
+      : await vscode.workspace.openTextDocument(); // opens an untitled document
```
yonip23, line 22: I don't think you need to do that, if `docUri` is `undefined`, calling `vscode.workspace.openTextDocument(docUri)` is equivalent to calling it without arguments.

src/test/suite/untitledFile.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    editor = editor as vscode.TextEditor;
+    doc = doc as vscode.TextDocument;
+    await editor.edit((edit) => {
+      edit.insert(new vscode.Position(0, 0), input);
+    });
+    await sleep(1000); // wait for editor to detect language
```
yonip23, line 28: isn't there any better way to confirm that the editor has detected the language? tests with `sleep`s are prone to being flaky.

src/test/suite/utils/helper.ts
yonip23: if `docUri` is undefined, it'd be considered the same as calling the function without any arguments --> the overload will resolve to the first one.

give it a try, i'm pretty sure it'll work.

src/test/suite/untitledFile.test.ts
yonip23: mmm ok then, lets hope it'll be stable enough 🙂

src/test/suite/utils/helper.ts
yonip23: oh well, then you can resolve this comment.
sorry for bothering you 🙈

yonip23 (approved):

src/test/suite/utils/helper.ts
assapir: this is why I hate TS :(

src/runCompletion.ts
```diff
@@ -48,3 +49,23 @@ function getMaxResults(): number {
...
+export function getLanguageFileExtension(
+  languageId: string
+): string | undefined {
+  return languages[languageId as KnownLanguageType];
+}
+
+export function getFileNameWithExtension(document: TextDocument): string {
+  const { languageId, fileName } = document;
+  let filenameWithExtension = fileName;
+  if (document.isUntitled) {
```
assapir, line 64: I always prefer early returns, so:
```suggestion
  if (!document.isUntitled) {
       return filenameWithExtension;
   }
```

dimacodota (approved):
lgtm

src/runCompletion.ts
```diff
@@ -48,3 +49,23 @@ function getMaxResults(): number {
...
+
+export function getLanguageFileExtension(
+  languageId: string
+): string | undefined {
+  return languages[languageId as KnownLanguageType];
+}
+
+export function getFileNameWithExtension(document: TextDocument): string {
+  const { languageId, fileName } = document;
+  let filenameWithExtension = fileName;
```
assapir, line 63: why is that needed?

src/runCompletion.ts
```diff
@@ -48,3 +49,23 @@ function getMaxResults(): number {
...
+  if (document.isUntitled) {
+    const extension = getLanguageFileExtension(languageId);
+    if (extension) {
+      filenameWithExtension = fileName.concat(extension);
```
assapir, line 67:
```suggestion
      return  fileName.concat(extension);
```

assapir (approved):

src/runCompletion.ts
LeshemNoa: Current binary API infers the file's language using its full name including the file extension. Untitled files have no file extension to their name. That way we can add the file extension manually to the untitled file's name.

assapir (approved):
Ahla

PR #881, opened 2022-07-06
Open welcome when init is done

ofekby:


src/extension.ts
```diff
@@ -134,14 +132,13 @@ async function backgroundInit(context: vscode.ExtensionContext) {
   setDefaultStatus();
   void registerCommands(context);
   pollDownloadProgress();
+  pollProcessState(() => {
+    void handleOpenWelcomeInHub(context);
+  });
```
yanir-codota, line 137: won't this open the welcome/hub to all users? how will we implement the AB test? is it the `should_open_welcome_in_hub` flag?

src/openWelcomeInHubFlag.ts
```diff
@@ -0,0 +1,17 @@
+import { ExtensionContext } from "./preRelease/types";
+
+const ALREADY_OPENED_WELCOME_FLAG = "already-opened-welcome";
+
+export async function setAlreadyInstalledFlag(
```
yanir-codota, line 5: I assume this to avoid opening the welcome/hub again at next restart, where is this called? what triggers this flag?

src/extension.ts
ofekby: No, the logic for opening the hub is in `should_open_welcome_in_hub` and yes it is using the flag from the binary

src/openWelcomeInHubFlag.ts
ofekby: Correct

PR #882, opened 2022-07-11
override tab key for inline

dimacodota:
**motivation**: in the case where both a popup and the inline are visible, 'tab` key selects the popup suggestion:

**solution**: 
this pr adds the following : 
 - `lookahead` mode(the code was extracted on moved to a new file) - in this mode tabnine extends the selected  popup item and suggests a longer inline suggestion
 - the current `lookahead` suggestion is stored
 - we override the `tab` key and check if there is an active `lookahead` suggestion and accept it if any and fallback to the regular acceptance logic if there is no active `lookahead` suggestion
 
 
**before this fix**:

https://user-images.githubusercontent.com/60742964/178226600-da3a9a43-d2af-4432-8d58-bbf07f4389bb.mov 

**after this fix**:

https://user-images.githubusercontent.com/60742964/178226594-18e62b6c-e39c-461e-b175-ab89724d5735.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -336,6 +336,11 @@
 				"when": "tabnine.snippet-suggestion:enabled && tabnine.in-inline-suggestions || tabnine.inline-suggestion:enabled && tabnine.in-inline-suggestions"
 			},
 			{
+				"key": "tab",
+				"command": "tabnine.tab-override",
+				"when": "tabnine.tab-override && suggestWidgetVisible && textInputFocus"
+			},
+			{
 				"key": "escape",
 				"command": "tabnine.escape-inline-suggestion",
 				"when": "tabnine.snippet-suggestion:enabled && tabnine.in-inline-suggestions || tabnine.inline-suggestion:enabled && tabnine.in-inline-suggestions"
--- a/src/lookAheadSuggestion.ts
+++ b/src/lookAheadSuggestion.ts
@@ -85,27 +85,21 @@ function registerTabOverride(): Disposable {
   return commands.registerTextEditorCommand(
     `${TAB_OVERRIDE_COMMAND}`,
     (_textEditor: TextEditor, edit: TextEditorEdit) => {
-      const { range, insertText } = currentLookAheadSuggestion ?? {
-        range: undefined,
-        insertTex: undefined,
-      };
+      if (!currentLookAheadSuggestion) {
+        void commands.executeCommand("acceptSelectedSuggestion");
+        return;
+      }
+
+      const { range, insertText } = currentLookAheadSuggestion;
       if (range && insertText) {
         edit.replace(range, insertText);
-      } else {
-        void commands.executeCommand("acceptSelectedSuggestion");
       }
     }
   );
 }
+
 async function enableTabOverrideContext(): Promise<Disposable> {
-  const res = await commands.executeCommand(
-    "setContext",
-    "tabnine.tab-override",
-    true
-  );
-
-  console.log("res after enableTabOverrideContext: ", res);
-
+  await commands.executeCommand("setContext", "tabnine.tab-override", true);
   return {
     dispose() {
       void commands.executeCommand(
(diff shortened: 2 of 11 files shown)


src/test/suite/completion.test.ts
```diff
@@ -44,4 +61,53 @@ describe("Should do completion", () => {
...
+  expected: string,
+  doBeforeInline?: () => void
+) {
+  await openDocument("javascript", "cons");
+  await isProcessReadyForTest();
+  await moveToActivePosition();
+  await makeAChange("o");
+  await emulationUserInteraction();
+  await vscode.commands.executeCommand("editor.action.triggerSuggest");
+  doBeforeInline?.();
```
yonip23, line 104: `?.()`
this is insane

src/test/suite/completion.test.ts
```diff
@@ -44,4 +61,53 @@ describe("Should do completion", () => {
...
+  it("should prefer an inline when both popup and inline are visible", async () => {
+    await assertSuggestionWith("console.log", () => {
+      mockAutocomplete(
+        requestResponseItems,
+        anAutocompleteResponse("console", "console.log")
+      );
+    });
+  });
 });
+async function assertSuggestionWith(
```
yonip23, line 94: too generic name, its not clear at all what does this function do. especially when there are tests which only calls this function and nothing else.

either separate it to more functions / rename it to be very descriptive

src/test/suite/completion.test.ts
```diff
@@ -44,4 +61,53 @@ describe("Should do completion", () => {
...
+    await assertSuggestionWith("console");
+  });
+  it("should prefer an inline when both popup and inline are visible", async () => {
+    await assertSuggestionWith("console.log", () => {
+      mockAutocomplete(
+        requestResponseItems,
+        anAutocompleteResponse("console", "console.log")
+      );
+    });
+  });
```
yonip23, line 92: its not clear what are these tests actually do. 
along with my other comment, i'd recommend extracting the logic to more functions

src/lookAheadSuggestion.ts
```diff
@@ -0,0 +1,118 @@
...
+  );
+}
+async function enableTabOverrideContext(): Promise<Disposable> {
+  const res = await commands.executeCommand(
+    "setContext",
+    "tabnine.tab-override",
+    true
+  );
+
+  console.log("res after enableTabOverrideContext: ", res);
```
yonip23, line 107: do you need this in production? if not, remove

src/lookAheadSuggestion.ts
```diff
@@ -0,0 +1,118 @@
...
+    (_textEditor: TextEditor, edit: TextEditorEdit) => {
+      const { range, insertText } = currentLookAheadSuggestion ?? {
+        range: undefined,
+        insertTex: undefined,
+      };
+      if (range && insertText) {
+        edit.replace(range, insertText);
+      } else {
+        void commands.executeCommand("acceptSelectedSuggestion");
+      }
```
yonip23, line 96: i think its more straight forward to write it something like:

```typescript
if (!currentLookAheadSuggestion) {
    void commands.executeCommand("acceptSelectedSuggestion");
    return;
}

const { range, insertText } = currentLookAheadSuggestion;
edit.replace(range, insertText);
```

package.json
```diff
@@ -335,6 +335,11 @@
 				"command": "tabnine.accept-inline-suggestion",
 				"when": "tabnine.snippet-suggestion:enabled && tabnine.in-inline-suggestions || tabnine.inline-suggestion:enabled && tabnine.in-inline-suggestions"
 			},
+			{
+				"key": "tab",
+				"command": "tabnine.tab-override",
+				"when": "tabnine.tab-override && suggestWidgetVisible && textInputFocus"
+			},
```
yonip23, line 342: how does it play with the other "tab" key we're defining? can there be a situation where both key overrides are active? what happens if so?

yonip23 (changes requested):
im requesting changes mainly for the tests, other than that great work 🙂

package.json
dimacodota: each override should be guarded with the relevant context flag. this override is enabled only if `tabnine.tab-override` context flag is enabled.

PR #901, opened 2022-07-27
First suggestion hint

LeshemNoa:
If the notification has not yet been shown, and the binary's `installation_time` is within the last hour, on the first suggestion we display the message for 5 seconds, and then it disappears. The message is anchored to the right of the first line of the suggestion. Currently looks like so for a single line suggestion
image
And for a multiline suggestion
image

Here's a video too 
https://user-images.githubusercontent.com/39261309/181248003-615d17e4-babe-4649-a861-2deacc0fd18b.mov

Would love to hear everybody's thought on
- The message text, style and look
- The behavior: is the timeout a good enough solution? do we really want to only display this message once?
- Anything else!


src/firstSuggestionHint.ts
```diff
@@ -0,0 +1,69 @@
...
+  
+  export function displayFirstSuggetionHint(position: Position) {
+    window.activeTextEditor?.setDecorations(
+      firstSuggestionDecorationType,
+      getFirstCompletionDecoration(position)
+    );
+    // for now: clear notification after 5 seconds
+    setTimeout(() => {
+        clearFirstSuggestionHint();
+    }, 10000);
```
yonip23, line 39: this doesn't look like 5s to me 🙂

src/firstSuggestionHint.ts
```diff
@@ -0,0 +1,69 @@
...
+import getHintColor from "./inlineSuggestions/hintColor";
+import { firstSuggestionHintAlreadyDisplayed, setFirstSuggestionFlag } from './firstSuggestionHintFlag';
+import { getState } from "./binary/requests/requests";
+import isInTheLastHour from "./utils/time.utils";
+import { getTabnineExtensionContext } from './globals/tabnineExtensionContext';
+
+// Anchor hint to the right of the line
+const firstSuggestionDecorationType = window.createTextEditorDecorationType({
+    rangeBehavior: DecorationRangeBehavior.ClosedOpen,
+});
```
yonip23, line 11: can be a function, no need to create this const globally.

src/firstSuggestionHint.ts
yonip23, line 39:
```suggestion
    setTimeout(clearFirstSuggestionHint, 10000);
```

src/firstSuggestionHintFlag.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const FIRST_SUGGESTION_HINT_SHOWN = "already-displayed-first-suggestion-hint";
+
+export async function setFirstSuggestionFlag(
+  context: ExtensionContext
+): Promise<void> {
+  await context.globalState.update(FIRST_SUGGESTION_HINT_SHOWN, true);
+}
+
+export async function unsetFirstSuggestionFlag(
```
yonip23, line 11: who's using this function?

src/firstSuggestionHint.ts
```diff
@@ -0,0 +1,69 @@
+import { DecorationOptions, DecorationRangeBehavior, ExtensionContext, Position, Range, window } from "vscode";
```
yonip23, line 1: the terminology here is a bit off imo - all references like `displayFirstSuggetionHint` should be `displayFirstSuggetionDecoration`

yonip23 (commented):
i'd like to see 1 or 2 basic tests for this functionality 🙂

yonip23:
> and the binary's installation_time is within the last hour

@LeshemNoa why do we need to assert that? what if the user didn't work with tabnine within the first hour?

yonip23 (approved):
Lgtm 👏

LeshemNoa:
@yonip23 Dima came up with this solution. It's the same approximation we use to display the welcome hub. This was seemingly the best workaround that we could do to capture this event at the plugin level (rather than doing it at the binary level, which Dima and Yair decided not to).

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,74 @@
...
+import { firstSuggestionDecorationAlreadyDisplayed, setFirstSuggestionFlag } from './firstSuggestionDecorationFlag';
+import { fireEvent, getState } from "./binary/requests/requests";
+import isInTheLastHour from "./utils/time.utils";
+import { getTabnineExtensionContext } from './globals/tabnineExtensionContext';
+
+// Anchor hint to the right of the line
+const firstSuggestionDecorationType = window.createTextEditorDecorationType({
+    rangeBehavior: DecorationRangeBehavior.ClosedOpen,
+});
+const maxSmallIntegerV8 = 2 ** 30; // Max number that can be stored in V8's smis (small integers)
```
yonip23, line 12: we've got a `consts` file somewhere, i think it can be moved to there (with all caps)

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,74 @@
...
+    decorations.push({
+      renderOptions: {
+        after: {
+          color: getHintColor(),
+          contentText: "👈 Press Tab to accept Tabnine's suggestion",
+          margin: "20px",
+          border: '0.5px solid',
+        },
+      },
+      range: document.validateRange(new Range(position.line, maxSmallIntegerV8, position.line, maxSmallIntegerV8)),
```
yonip23, line 28: why do you need to validate? i think its very unusual to see a file with 2^30 lines...

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,74 @@
...
+// message should be displayed if it has not been displayed yet, 
+// and if the extension has been installed within the last hour 
+export async function shouldDisplayFirstSuggestionDecoration(
+    context: ExtensionContext
+      ): Promise<boolean> {
+    if (firstSuggestionDecorationAlreadyDisplayed(context)) {
+      return false;
+    }
+    const state = await getState();
+    return Boolean(
```
yonip23, line 55: y like dis? why not using `!!` if you want to convert to boolean?

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+export async function handleFirstSuggestionDecoration(position: Position): Promise<void> {
+    const context = getTabnineExtensionContext();
+    if (context && await shouldDisplayFirstSuggestionDecoration(context)) {
+        displayFirstSuggestionDecoration(position);
+        await fireEvent({
+          name: "first-suggestion-decoration-displayed",
+        });
+        await setFirstSuggestionFlag(context);
```
yonip23, line 68: not sure if critical, but both `await`s can be combined to `Promise.all`

src/provideInlineCompletionItems.ts
```diff
@@ -26,12 +27,16 @@ export default async function provideInlineCompletionItems(
...
     if (completionInfo) {
       return await getLookAheadSuggestion(document, completionInfo, position);
     }
-
-    return await getInlineCompletionItems(document, position);
+    const completions = await getInlineCompletionItems(document, position);
+    if (completions.items.length > 0 && !isEmptyCompletion(completions.items[0])) {
+      await handleFirstSuggestionDecoration(position);
+    }
+    return completions;
```
yonip23, line 39: I'd move the `if` to inside the `handle`, so that the caller code (i.e this code) will not have to be aware of the business logic of the "first suggestion decoration" feature:

```suggestion
const completions = await getInlineCompletionItems(document, position);
await handleFirstSuggestionDecoration(completion, position);
return completions;
```

src/firstSuggestionDecoration.ts
LeshemNoa: What `validateRange`does is "Ensure a range is completely contained in this document". This trick allows me to calculate the range that'll allow me to anchor the banner to the end of the line. I took this from [Gitlens](https://github.com/gitkraken/vscode-gitlens/blob/e916ecb0d14bed5baa9092d37c1a8569b6a1d276/src/annotations/lineAnnotationController.ts#L313). It was necessary to fix a bug I had otherwise - sometimes the banner location and typing into the editor would somehow race and cause the banner to be rendered in the middle of the line with the suggestion on the right. This solved it.

yonip23: legit, well done for catching it 🔥

src/test/suite/checkFirstSuggestionDecoration.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  INLINE_NEW_PREFIX,
+  INLINE_PREFIX,
+  SINGLE_CHANGE_CHARACTER,
+} from "./utils/testData";
+import * as vscodeUtils from "../../vscode.api";
+import { firstSuggestionDecorationType } from "../../firstSuggestionDecoration";
+
+const nonEmptyArray = sinon.match((v) => Array.isArray(v) && v.length);
+
+suite("First suggestion decoration", () => {
```
LeshemNoa, line 35: I had to give this file a weird name in order for it to run before the `fetchBinaryPath` test. Because all the tests run in the same extension host they may interfere with one another's environment, and `fetchBinaryPath` does something that makes this new test fail. I think we should put in place something like a super suite, that'll allow us to determine the order in which the tests will run manually, rather than having them just run in alphabetical order. @dimacodota

src/vscode.api.ts
```diff
@@ -0,0 +1,18 @@
...
+  window,
+} from "vscode";
+
+// eslint-disable-next-line import/prefer-default-export
+export const setDecoration = (
+  decorationType: TextEditorDecorationType,
+  rangesOrOptions: Range[] | DecorationOptions[],
+  editor?: TextEditor
+): void =>
+  (editor ?? window.activeTextEditor)?.setDecorations(
```
LeshemNoa, line 15: This is the hack you implemented, @dimacodota, but I made a small change to it. 
Consider the following situation: we display a decorator in one editor tab, while it's displayed the user switches tabs. Now we wish to clear the decorator. 
If we don't take the editor as a parameter and just use the currently active editor tab, then we will clear the decorator on the currently active tab and not on the tab where it was actually displayed.

src/test/suite/checkFirstSuggestionDecoration.test.ts
yonip23: In general, if tests are sensitive to execution order they're not done right. 
I think we should look into this issue and understand exactly what is the link between the 2 tests, and break it.

LeshemNoa: Besides while I was learning the ins and outs of these (highly complicated!!) tests, I found that there's some tests that are very flaky in the CI. I had some issues particularly with the notification polling test but I couldn't pin it down.

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,120 @@
...
+import { MAX_SMALL_INTEGER_V8 } from "./globals/consts";
+import { Capability, isCapabilityEnabled } from "./capabilities/capabilities";
+// Anchor hint to the right of the line
+export const firstSuggestionDecorationType = window.createTextEditorDecorationType(
+  {
+    rangeBehavior: DecorationRangeBehavior.ClosedOpen,
+  }
+);
+
+export function getFirstCompletionDecoration(
```
dimacodota, line 30: `firstSuggestion`?

dimacodota: why it is exported?

src/test/suite/checkFirstSuggestionDecoration.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (args[0] === firstSuggestionDecorationType) {
+          if (args[1].length === 0) {
+            events.emit("cleared");
+          } else {
+            events.emit("shown");
+          }
+        }
+        original(...args);
+      });
+    const decorationShown = new Promise((res) => events.once("shown", res));
```
dimacodota, line 68: not sure I understand this, once is promise based already

src/test/suite/utils/completion.utils.ts
```diff
@@ -65,6 +71,22 @@ export function mockAutocomplete(
     result,
   });
 }
+
+export async function mockGetState(
```
dimacodota, line 75: would move it to a different file, it is not related to completions

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,120 @@
...
+  if (!editor) {
+    return;
+  }
+  setDecoration(
+    firstSuggestionDecorationType,
+    getFirstCompletionDecoration(editor.document, position),
+    editor
+  );
+  // for now: clear notification after 5 seconds
+  setTimeout(clearFirstSuggestionDecoration, 5000, editor);
```
dimacodota, line 67: i think you need to clear it on the selection change event as well. What would happen if you press the down-arrow key while the hint is shown?

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,120 @@
...
+  completions: InlineCompletionList
+): Promise<void> {
+  if (
+    completions.items.length === 0 ||
+    isEmptyCompletion(completions.items[0])
+  ) {
+    return;
+  }
+  const context = getTabnineExtensionContext();
+  if (context && (await shouldDisplayFirstSuggestionDecoration(context))) {
```
dimacodota, line 107: maybe you can simplify this by passing `context?` and  `return !!context?.globalState.get<boolean>` in firstSuggestionDecorationAlreadyDisplayed

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,120 @@
...
+    displayFirstSuggestionDecoration(position);
+    await Promise.all([
```
dimacodota, line 109: i think you definitely want to `await setFirstSuggestionFlag` but probably `void fireEvent`...

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,120 @@
...
+    rangeBehavior: DecorationRangeBehavior.ClosedOpen,
+  }
+);
+
+export function getFirstCompletionDecoration(
+  document: TextDocument,
+  position: Position
+): DecorationOptions[] {
+  const decorations: DecorationOptions[] = [];
+  decorations.push({
```
dimacodota, line 35: why an array? just return the `DecorationOptions`

src/test/suite/checkFirstSuggestionDecoration.test.ts
LeshemNoa: Node's `EventEmitter.once` is only a hook for a callback to do when an event happens. It's not promise based ([see here](https://github.com/nodejs/node/issues/20893)). [Emittery's `once`](https://www.npmjs.com/package/emittery) does return a promise though, but it requires adding a dependency.

src/firstSuggestionDecoration.ts
LeshemNoa: The installation time is not in the globalState, it's in the binary state.

src/firstSuggestionDecoration.ts
LeshemNoa: You're right, arrow down leaves the decorator in its place. However, the `window.onDidChangeTextEditorSelection` handler doesn't fire at all when you have an inline suggestion shown and you press the down arrow, since this seems to be handled entirely by the inline completion provider API...

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,125 @@
...
+        position.line,
+        MAX_SMALL_INTEGER_V8,
+        position.line,
+        MAX_SMALL_INTEGER_V8
+      )
+    ),
+  };
+}
+
+export function displayFirstSuggestionDecoration(position: Position): void {
```
dimacodota, line 55: why exported?

src/firstSuggestionDecoration.ts
```diff
@@ -0,0 +1,125 @@
...
+    [getFirstSuggestionDecoration(editor.document, position)],
+    editor
+  );
+  // for now: clear notification after 5 seconds
+  setTimeout(clearFirstSuggestionDecoration, 5000, editor);
+}
+
+// message should be displayed if it has not been displayed yet,
+// and if the extension has been installed within the last hour
+export async function shouldDisplayFirstSuggestionDecoration(
```
dimacodota, line 71: why exported?

dimacodota (approved):
lgtm

PR #902, opened 2022-07-27
attribute inline suggestion

dimacodota:
this feature is enabled only for `ATTRIBUTION`+ `ALPHA_CAPABILITY` capabilities

attribution


dimacodota:
Since mocha tests are executed in alphabetical order, `capabilities.test.ts`  which tests that capabilities are fetched when the extension is activated and must be executed before all others was renamed to `0 - capabilities.test.ts`. ref: https://stackoverflow.com/a/39064345

src/vscode.api.ts
```diff
@@ -0,0 +1,13 @@
+import {
```
dimacodota, line 1: since there is no way to spy on `setDecorations` because it is a non-enumerable property, this wrapper was introduced.

src/capabilities/capabilities.ts
```diff
@@ -31,13 +31,17 @@ export enum Capability {
...
+  ATTRIBUTION = "attribution",
 }
 
 let enabledCapabilities: Record<string, boolean> = {};
 
 export function isCapabilityEnabled(capability: Capability): boolean {
   return enabledCapabilities[capability];
 }
+export function isAnyCapabilityEnabled(...capabilities: Capability[]): boolean {
+  return capabilities.some((capability) => !!enabledCapabilities[capability]);
```
yonip23, line 43: Why not use `includes`?

src/test/suite/attribute.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import {
+  anAutocompleteResponse,
+  INLINE_NEW_PREFIX,
+  INLINE_PREFIX,
+  SINGLE_CHANGE_CHARACTER,
+} from "./utils/testData";
+import * as vscodeApi from "../../vscode.api";
+
+describe("Should show attribution item on inline suggestion", () => {
+  const docUri = getDocUri("completion.txt");
```
yonip23, line 35: Maybe create another file for this test? I don't think it's a good idea sharing files between tests

src/test/suite/attribute.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      requestResponseItems,
+      anAutocompleteResponse(INLINE_PREFIX, INLINE_NEW_PREFIX)
+    );
+    await sleep(1000);
+    await moveToActivePosition();
+    await makeAChange(SINGLE_CHANGE_CHARACTER);
+    await triggerInline();
+
+    await sleep(1000);
+    assert(
```
yonip23, line 81: Needless assert - it is being tested in the previous test

src/vscode.api.ts
yonip23: I don't like the idea of making such a change just for tests - Isn't there an api to get current decorations from the editor? If it exists, maybe you can use it in the tests to assert that our decoration exists?

yonip23 (commented):
Great work, well done!
See comments

src/attribute.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export default function attribute(position: Position): void {
+  if (
+    isAnyCapabilityEnabled(Capability.ATTRIBUTION, Capability.ALPHA_CAPABILITY)
+  ) {
+    drawDecorationAtCurrentLine(position);
+  }
+}
+
+window.onDidChangeTextEditorSelection(clearDecoration);
```
amirbilu, line 24: This is a bit side effect-ish. no?

src/attribute.ts
```diff
@@ -0,0 +1,32 @@
...
+import { setDecoration } from "./vscode.api";
+
+const logoPath = Uri.file(path.resolve(__dirname, "..", "small_logo-mono.png"));
+
+const type = window.createTextEditorDecorationType({
+  gutterIconPath: logoPath,
+  gutterIconSize: "70%",
+});
+
+export default function attribute(position: Position): void {
```
amirbilu, line 16: I think this module deserves a bit more descriptive name + consider move this under `inlineSuggestions`

src/capabilities/capabilities.ts
dimacodota: because I want at least one of the input capabilities to be in `enabledCapabilities`

src/attribute.ts
dimacodota: not a bit, it is a side effect. but what is your question /suggestion here?

amirbilu: Add it under method and call it somewhere..

src/vscode.api.ts
dimacodota: Well, I didn't like it either. obviously, the existence of such an API would make life much easier, but unfortunately, there is none (to remind you that we were together on a quest to find  one where you implemented the first iteration of the inline)

PR #903, opened 2022-07-28
remove inline hint color settings

yonip23:
since we're using vscode's inline completions API we no longer control the color of the hint - so this settings property has no effect on the users.


assapir (approved):

talcodota:
:(

PR #907, opened 2022-07-28
readme page textual content and images updated, emojis removed from e…

YotamFromCodota:
…xtension description in package.json


assapir (approved):

assapir (commented):
Change the Twitter handle also?

PR #913, opened 2022-08-03
Support hub tabnineUrl query param

ofekby:


src/hub/hubUri.ts
```diff
@@ -12,17 +15,19 @@ export default async function hubUri(
...
+  const tabnineUrl = hubUrl.searchParams.get(TABNINE_URL_QUERY_PARAM);
+  if (tabnineUrl) {
+    hubUrl.searchParams.set(
+      TABNINE_URL_QUERY_PARAM,
+      (await env.asExternalUri(Uri.parse(tabnineUrl))).toString()
+    );
+  }
+
+  if (path) {
+    hubUrl.pathname += path.startsWith("/") ? path : `/${path}`;
```
amirbilu, line 29: why not using `joinPath`?

src/hub/hubUri.ts
```diff
@@ -1,6 +1,9 @@
-// import { URL } from "url";
+import { URL } from "url";
 import { Uri, env } from "vscode";
-import { StateType /* TABNINE_URL_QUERY_PARAM */ } from "../globals/consts";
+import {
+  StateType /* TABNINE_URL_QUERY_PARAM */,
```
assapir, line 4:
```suggestion
  StateType,
```

src/hub/hubUri.ts
```diff
@@ -12,17 +15,19 @@ export default async function hubUri(
...
-  //     TABNINE_URL_QUERY_PARAM,
-  //     (await env.asExternalUri(Uri.parse(tabnineUrl))).toString()
-  //   );
-  // let uri = Uri.parse(hubUrl.toString());
-  //
-  if (path) uri = Uri.joinPath(uri, path);
-  return uri;
+  const tabnineUrl = hubUrl.searchParams.get(TABNINE_URL_QUERY_PARAM);
+  if (tabnineUrl) {
+    hubUrl.searchParams.set(
```
assapir, line 22: you are setting it if it exist, not if not exist?

src/hub/hubUri.ts
ofekby: It since I preferred using the URL object and not the URI which doesn't support query params that well.

src/hub/hubUri.ts
ofekby: Correct it just fixing the it to be external URI

PR #923, opened 2022-08-14
Fix wrong path join of paths in hub

ofekby:


src/hub/hubUri.ts
```diff
@@ -22,8 +23,8 @@ export default async function hubUri(
     );
   }
 
-  if (path) {
-    hubUrl.pathname += path.startsWith("/") ? path : `/${path}`;
+  if (hubPath) {
+    hubUrl.pathname = path.join(hubUrl.pathname, hubPath);
```
assapir, line 27: That url path or fs path? If it's url, I think it's not the right lib to use

ofekby: Path is a path is a path

assapir: nope https://stackoverflow.com/a/16301947/5520074

assapir: Because it won't work as expected in windows for example

ofekby: This is not URL, it is a join between two paths, hubUrl.pathname and hubPath

assapir: You are joining url paths. The path separator is \ in windows, so it will be miss behaving

ofekby: Never mind the VSCODE extensions can't use this package anyway

assapir (approved):

PR #926, opened 2022-08-17
support code server asExternalUrl - to open the hub correctly

amirbilu:


amirbilu:
image

src/globals/consts.ts
```diff
@@ -157,3 +157,4 @@ export const TABNINE_APP_URL = "https://app.tabnine.com";
 export const URI_SCHEME_FILE = "file";
 export const BINARY_RESTART_EVENT = "binary-restart-event";
 export const MAX_SMALL_INTEGER_V8 = 2 ** 30; // Max number that can be stored in V8's smis (small integers)
+export const LOCAL_ADDRESSES = ["localhost", "127.0.0.1"];
```
assapir, line 160: 0.0.0.0?🥲

assapir (approved):
Boba

src/globals/consts.ts
amirbilu: Tachles.. we format the address.. and we hard-coded to 127.0.0.1

PR #927, opened 2022-08-21
fix the selected text when an inline suggestion is selected and the popup is visible

dimacodota:
The issue:
In the case where both inline and popup are visible and inline is accepted, the accepted text is marked as selected after selection, so any change will remove the selected area

before:
selection-before

after:
selection-after

### DIFF
--- a/src/test/suite/completion.test.ts
+++ b/src/test/suite/completion.test.ts
@@ -104,6 +104,10 @@ describe("Should do completion", () => {
     await triggerSelectionAppetence();
 
     assertTextIsCommitted("console.log");
+
+    await makeAChange("o");
+
+    assertTextIsCommitted("console.logo");
   });
 });
 
--- a/src/test/suite/completion.test.ts
+++ b/src/test/suite/completion.test.ts
@@ -105,11 +105,15 @@ describe("Should do completion", () => {
 
     assertTextIsCommitted("console.log");
 
-    await makeAChange("o");
-
-    assertTextIsCommitted("console.logo");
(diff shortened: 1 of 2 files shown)


src/test/suite/completion.test.ts
```diff
@@ -104,9 +104,17 @@ describe("Should do completion", () => {
     await triggerSelectionAppetence();
 
     assertTextIsCommitted("console.log");
+
+    await makeAndAssertFollowingChange();
```
yonip23, line 108: I don't quite understand how is this testing the behavior relevant to this pr

yonip23 (approved):
Lgtm

src/test/suite/completion.test.ts
dimacodota: it is making an additional change after the inline suggestion was accepted, and asserting the accepted text + current change are existing. the issue was that if you type something after accepting an inline suggestion, the last change would replace the selected inline suggestion.

yonip23: So I'd extract it to another test with a proper name (and maybe comments inside the test to explain how it is testing this behavior)

Testing multiple things in one test is generally not a good idea, as it makes it harder to find what went wrong if the test fails

yonip23: Also, this sounds a bit vicarious. Isn't there a way to get the current selected text from the editor and assert that it is empty / null?

dimacodota: agree, will do

PR #945, opened 2022-09-28
fix lookAhead dot but

yairco1990:
fix for the following use case:
- user is typing dot (".")
- vscode shows completions (inside the popup) that don't start with "."
- the user focuses on one of those suggestions
- the SelectedCompletionInfo text contains that dot at the beginning of the completion
- we are filtering completions that have the same prefix as the SelectedCompletionInfo text, but don't find any. (since "use" != ".use")
- the user doesn't get any completion

image (1)
image

the fix is to pass the SelectedCompletionInfo text to the filter function without the dot

### DIFF
--- a/src/lookAheadSuggestion.ts
+++ b/src/lookAheadSuggestion.ts
@@ -49,10 +49,7 @@ export async function getLookAheadSuggestion(
     2
   );
 
-  const result = findMostRelevantSuggestion(
-    response,
-    getCompletionInfoWithoutOverlappingDot(completionInfo)
-  );
+  const result = findMostRelevantSuggestion(response, completionInfo);
   const completion =
     result &&
     response &&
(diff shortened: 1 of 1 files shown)


src/lookAheadSuggestion.ts
```diff
@@ -49,7 +49,10 @@ export async function getLookAheadSuggestion(
     2
   );
 
-  const result = findMostRelevantSuggestion(response, completionInfo);
+  const result = findMostRelevantSuggestion(
+    response,
+    getCompletionInfoWithoutOverlappingDot(completionInfo)
```
yonip23, line 54: i think its more concise to pass the `completionInfo` object here as is, and call `getCompletionInfoWithoutOverlappingDot` inside `findMostRelevantSuggestion` - as this logic is relevant only to this function.

yonip23 (approved):
lgtm - see comment

PR #956, opened 2022-10-18
disable suggestion on Tab/selection

dimacodota:
this pr fixes two issues:

 - disable suggestions on suggestion selection(Tab) or paste action 
 - disables suggestions on Tab key (indent in)


src/inlineSuggestions/stateTracker.ts
```diff
@@ -34,3 +43,21 @@ export function initTracker(): Disposable[] {
...
+  }
+
+  isSingleCharChange() {
+    return this.contentChange?.text.trim().length <= 1;
+  }
+
+  isNotIndentationChange() {
+    return this.contentChange?.text !== " ".repeat(getTabSize());
+  }
+}
```
yonip23, line 63: rename to something more specific (`TabnineDocumentChange` for example) and extract to separate file

src/inlineSuggestions/stateTracker.ts
```diff
@@ -23,9 +30,11 @@ export function initTracker(): Disposable[] {
   return [
     workspace.onDidChangeTextDocument(
       ({ contentChanges }: TextDocumentChangeEvent) => {
-        const contentChange = contentChanges[0];
+        const contentChange = new Change(contentChanges[0]);
         const changeHappened =
-          contentChange?.rangeLength >= 0 && contentChange?.text !== "";
+          contentChange.isValid() &&
+          contentChange.isNotIndentationChange() &&
+          contentChange.isSingleCharChange();
```
yonip23, line 37: there's a problem with this condition:
- in `isValid` you're assetring that `this.contentChange?.text.trim().length >= 0`
- in `isSingleCharChange` you're asserting that `this.contentChange?.text.trim().length <= 1`
==> you're asserting that `this.contentChange?.text.trim().length == 1`

this means that `isNotIndentationChange` will always evaluate to true, since it asserts that the text is not <a string that's length is `tabSize`> - but you already know that this string's length is exactly 1.
I'm not sure what you wanted to check here, I presume it should be only `isSingleCharChange`, or only `isNotIndentationChange` 🤷🏻‍♂️

src/test/suite/completion.test.ts
```diff
@@ -84,47 +95,107 @@ describe("Should do completion", () => {
...
     assertTextIsCommitted("console.log");
 
     await makeAndAssertFollowingChange();
   });
   it("should skip completion request on midline invalid position", async () => {
-    await openADocWith(" s");
+    await openADocWith("console s");
 
-    await moveToStartOfLinePosition();
+    await moveLeftBy(2);
```
yonip23, line 132: this test is not clear to me, because I don't know where is the cursor to begin with - so I don't know what's the meaning of moving it by 2

src/test/suite/completion.test.ts
```diff
@@ -84,47 +95,107 @@ describe("Should do completion", () => {
...
+    await emulationUserInteraction();
+
+    verify(
+      stdinMock.write(new SimpleAutocompleteRequestMatcher(), "utf8")
+    ).never();
+  });
+  it("should request completion on midline valid position", async () => {
+    await openADocWith("consol })");
+
+    await moveLeftBy(3);
```
yonip23, line 145: same here

yonip23 (changes requested):
overall looks good, please refer to my comments

src/inlineSuggestions/stateTracker.ts
dimacodota: not exactly...
the `isValid` assets untrimmed value - meaning it was an actual change (and not deletion for example)
the `isSingleCharChange` assert that the "timed" length is 1  - meaning it will allow any single not whitespace char
the `isNotIndentationChange` asserts a non-trimmed value - allowing everything different from the indent

I will rename the function names to better express the intention

yonip23: oh so you allow either [at least one non-whitespace char] or [only whitespace chars, but exactly `tabSize` length]
yeah that's a sneaky one - maybe worth playing around with the functions and conditions to make it clearer 👍🏻

yonip23 (approved):
thanks for replying - after your explanation lgtm

src/test/suite/completion.test.ts
```diff
@@ -84,47 +89,107 @@ describe("Should do completion", () => {
...
+    await emulationUserInteraction();
+
+    verify(
+      stdinMock.write(new SimpleAutocompleteRequestMatcher(), "utf8")
+    ).never();
+  });
+  it("should request completion on midline valid position", async () => {
+    await openADocWith("consol })");
+
+    await moveCursorToBeAfter("consol");
```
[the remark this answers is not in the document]
yonip23: Not bad, now I do understand what it does 😀
I think it would be even better to give an absolute position, something like

```typescript 
moveCursorFromStart(6);
```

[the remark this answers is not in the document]
yonip23: But not critical - you can leave it this way as well

PR #960, opened 2022-10-31
filter the relevant change by the current position

dimacodota:
**Problem:** 
in some cases, the onDidChangeTextDocument callback is raised with multiple contentChanges causing the current change to be filtered and thus not triggering any request in this position.

**Solution:** 
the array should be filtered by the current cursor position.

Screen Shot 2022-10-31 at 8 47 43

### DIFF
--- a/src/inlineSuggestions/documentChangesTracker/index.ts
+++ b/src/inlineSuggestions/documentChangesTracker/index.ts
@@ -24,7 +24,11 @@ export function initTracker(): Disposable[] {
   return [
     workspace.onDidChangeTextDocument(
       ({ contentChanges }: TextDocumentChangeEvent) => {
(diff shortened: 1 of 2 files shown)


src/inlineSuggestions/documentChangesTracker/index.ts
```diff
@@ -24,7 +24,11 @@ export function initTracker(): Disposable[] {
   return [
     workspace.onDidChangeTextDocument(
       ({ contentChanges }: TextDocumentChangeEvent) => {
-        const contentChange = new DocumentTextChangeContent(contentChanges[0]);
+        const currentPosition = window.activeTextEditor?.selection.active;
+        const relevantChange = contentChanges.find(
```
dimacodota, line 28: this is the most relevant change

amirbilu (approved):
lgtm

PR #966, opened 2022-11-08
Dev2-1649 wrong indentation in case of multine old suffix

dimacodota:
**problem:** 
in some cases, the binary  returns multiline `old_suffix` which is probably another issue that should be investigated

Screen Shot 2022-11-08 at 9 59 31

the solution in this pr is to modify the [replace range](https://github.com/codota/tabnine-vscode/blob/a12abaf4d511c67d1af69bdc9692995a12c0619f/src/vscode.proposed.inlineCompletions.d.ts#L105) only in case of a single  line which  should handle closing brackets that should be replaced


yairco1990 (approved):
lgtm

PR #967, opened 2022-11-10
DEV2-1677 check for tab-in in languages with tab indentation vs. spaces

dimacodota:
**problem:**
in the case of "tab" indentation (`\t`), the change detection was expecting only spaces `" "` and not tab indentation

### DIFF
--- a/src/binary/requests/tabSize.ts
+++ b/src/binary/requests/tabSize.ts
@@ -7,3 +7,6 @@ export default function getTabSize(): number {
   }
   return tabSize;
 }
+export function getTabsCount(): number {
+  return getTabSize() / getTabSize();
+}
--- a/src/binary/requests/tabSize.ts
+++ b/src/binary/requests/tabSize.ts
@@ -8,5 +8,5 @@ export default function getTabSize(): number {
   return tabSize;
 }
 export function getTabsCount(): number {
-  return getTabSize() / getTabSize();
+  return getTabSize() / 4;
 }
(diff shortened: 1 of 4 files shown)


src/binary/requests/tabSize.ts
```diff
@@ -7,3 +7,6 @@ export default function getTabSize(): number {
   }
   return tabSize;
 }
+export function getTabsCount(): number {
+  return getTabSize() / getTabSize();
```
yairco1990, line 11: what am I missing? isn't that just 1?

src/test/suite/completion.test.ts
```diff
@@ -178,19 +178,10 @@ describe("Should do completion", () => {
...
-    await triggerInline();
-
-    await emulationUserInteraction();
-
-    verify(
-      stdinMock.write(new SimpleAutocompleteRequestMatcher(), "utf8")
-    ).never();
+    await runSkipIndentInTest("javascript");
+  });
+  it("should skip completion request on Tab key (indention in) in golang where indentation is \t", async () => {
```
yairco1990, line 183: what's the difference between golang and other langs?

src/binary/requests/tabSize.ts
dimacodota: my bad 🤦‍♂️

src/test/suite/completion.test.ts
dimacodota: it just will use '\t' by default, it was easier than re-config the IDE indentation in the test

yairco1990 (approved):

PR #976, opened 2022-11-20
DEV2-1736 apply python indentation extension fix

dimacodota:
**Problem:**
if [vsc-python-indent](https://github.com/kbrose/vsc-python-indent)  is installed, this extension  is  overrides the "enter" key biding and creates an `edit.insert`   https://github.com/kbrose/vsc-python-indent/blob/47a0c164f44fce0d82f6c7f16fd844626914273f/src/indent.ts#L54

for some reason, if the edit starts with `\n` the provideInlineCompletionItems handler is not triggered (probably due to some vscode issue).

**Solution:** as a workaround in this case we will check if the edit is in a python file and the edit starts with `\n` and `vsc-python-indent` is enabled,  we will trigger the `provideInlineCompletionItems`


yairco1990 (approved):
lgtm

PR #978, opened 2022-11-21
DEV2-1254 do not change the order of the suggestions

dimacodota:
the order is handled by Tabnine engine(binary), the returned suggestions are ordered by the probability + length, so there is no need to reorder them, the only thing that should be handled is prefix matching.

ref https://tabnine.atlassian.net/browse/DEV2-1254


yairco1990 (approved):
lgtm

PR #980, opened 2022-11-23
DEV2-1763  - accept suggestion  with tab indentation  is not working 

dimacodota:
**problem:**
vscode's builtin  `editor.action.inlineSuggest.commit` is not triggered if the current suggestion has an indentation greater than `tab`https://github.com/microsoft/vscode/blob/45b5568e7d351bb75f2930f48b1978a4df016b84/src/vs/editor/contrib/inlineCompletions/browser/ghostTextController.ts

in our case there are suggestions with indentation that can greater than tab :

Screen Shot 2022-11-23 at 18 16 12

**Solution:** override the 'tab' keyword and trigger the builtin `editor.action.inlineSuggest.commit` command

### DIFF
--- a/package.json
+++ b/package.json
@@ -337,6 +337,11 @@
 				"when": "tabnine.tab-override && suggestWidgetVisible && textInputFocus"
 			},
 			{
+				"key": "tab",
+				"command": "editor.action.inlineSuggest.commit",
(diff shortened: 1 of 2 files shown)


yairco1990 (approved):
lgtm

PR #989, opened 2022-12-01
DEV2-1812

dimacodota:
this pr handles 2 issues:
 - support suggestion if the current change contains auto-closed  brackets 
 -  do not query tabnine if the focused popup item  is not equal to the prefix (see the screenshot):

Screen Shot 2022-12-01 at 18 02 20


amircodota (approved):
LGTM

PR #993, opened 2022-12-12
DEV2-1533 - introduce inline completions debounce 

dimacodota:
DEV2-1533

this pr adds a debounce before rendering capability as well as some tests refactoring

the debounce value can be modified via `debounceMilliseconds` property


src/autocompleteInstaller.ts
```diff
@@ -87,11 +86,11 @@ class InstallOptions {
     );
   }
 
-  public static get(testMode: boolean) {
+  public static get() {
```
dimacodota, line 89: i've added  a mockAutocompleteAPI test function, so there is  not need to check the test mode here

src/getInlineCompletionItems.ts
```diff
@@ -0,0 +1,54 @@
...
+import TabnineInlineCompletionItem from "./inlineSuggestions/tabnineInlineCompletionItem";
+import runCompletion from "./runCompletion";
+import getAutoImportCommand from "./getAutoImportCommand";
+import { SuggestionTrigger } from "./globals/consts";
+import { AutocompleteResult, ResultEntry } from "./binary/requests/requests";
+import { isMultiline } from "./utils/utils";
+
+export const INLINE_REQUEST_TIMEOUT = 3000;
+
+export default async function getInlineCompletionItems(
```
dimacodota, line 11: mostly refactoring, this function was extracted from [provideInlineCompletionItems.ts](https://github.com/codota/tabnine-vscode/pull/993/files#diff-88c55a307a2aafd4bd44fa02d2ceb1a8fffa4246cde36e7cf4981e2df04b8e75)

src/tryGetDebouncedCompletions.ts
```diff
@@ -0,0 +1,61 @@
+import * as vscode from "vscode";
+import { Capability, isCapabilityEnabled } from "./capabilities/capabilities";
+import getInlineCompletionItems from "./getInlineCompletionItems";
+import TabnineInlineCompletionItem from "./inlineSuggestions/tabnineInlineCompletionItem";
+import { sleep, timed } from "./utils/utils";
+
+export default async function tryGetDebouncedCompletions(
```
dimacodota, line 7: this is the main change

src/tryGetDebouncedCompletions.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function getDebounceValue(): number {
+  const debounceMilliseconds = vscode.workspace
+    .getConfiguration()
+    .get<number>("tabnine.debounceMilliseconds");
+  const isAlphaCapabilityEnabled = isCapabilityEnabled(
+    Capability.ALPHA_CAPABILITY
+  );
+  const ALPHA_ONE_SECOND_DEBOUNCE = 1000;
```
yonip23, line 56: move up to head of the file

src/tryGetDebouncedCompletions.ts
```diff
@@ -0,0 +1,61 @@
...
+  debounceTime: number
+) {
+  const canceledPromise = new Promise<void>((resolve) =>
+    token.onCancellationRequested(resolve)
+  );
+
+  await Promise.race([canceledPromise, sleep(debounceTime)]);
+}
+
+function calculateDebounceValue(time: number) {
```
yonip23, line 43:
```suggestion
function calculateDebounceMs(time: number) {
```
(also in `getDebounceValue` -> `Ms`)

src/tryGetDebouncedCompletions.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  if (debounceTime > 0) {
+    await debounceOrCancelOnRequest(token, debounceTime);
+
+    if (token.isCancellationRequested) {
+      return undefined;
+    }
+
+    // re fetch the most updated suggestions
+    return getInlineCompletionItems(document, position);
```
yonip23, line 28: I think this is excessive - how about checking the offset (or something similar to JB's `modificationStamp`) before and after sleeping, and re-fetching only if necessary?

src/tryGetDebouncedCompletions.ts
```diff
@@ -0,0 +1,61 @@
...
+): Promise<
+  vscode.InlineCompletionList<TabnineInlineCompletionItem> | undefined
+> {
+  const { time, value: current } = await timed(() =>
+    getInlineCompletionItems(document, position)
+  );
+
+  const debounceTime = calculateDebounceValue(time);
+
+  if (debounceTime > 0) {
```
yonip23, line 20: i'd invert the condition for simplicity, because the `else` branch is only a `return` statement

src/tryGetDebouncedCompletions.ts
yonip23, line 7: rename to `debounceCompletions` - the `try` is confusing imo

src/test/suite/completion.test.ts
```diff
@@ -39,14 +39,47 @@ import { resetBinaryForTesting } from "../../binary/requests/requests";
...
+  let sandbox: sinon.SinonSandbox;
+  let modeMock: sinon.SinonStub;
+
+  async function mockAutocompleteAPI() {
+    sandbox.reset();
+
+    modeMock?.returns(suggestionMode.SuggestionsMode.AUTOCOMPLETE);
+    await fetchCapabilitiesOnFocus();
+  }
+  function mockGetDebounceConfig(debounceValue: number): void {
```
yonip23, line 60: can these 2 functions be extracted to a util/driver?

src/test/suite/completion.test.ts
```diff
@@ -316,6 +345,38 @@ describe("Should do completion", () => {
       stdinMock.write(new SimpleAutocompleteRequestMatcher("return"), "utf8")
     ).never();
   });
+  it("should render suggestion after debounce", async () => {
```
yonip23, line 348: please add a test to the cancellation logic:
1. set long debounce
2. request 2 different completions requests with short sleep between them
3. assert request is fired once (the second request)

yonip23 (commented):
looking good! had some comments, but nothing too serious.

src/tryGetDebouncedCompletions.ts
yonip23: also, can this be a generic debounce function? something like
```typescript
function debounce<T>(
  action: () => Promise<T>,
  debounceMs: number,
  cancellation: () => Promise<boolean>
```

or something similar

package.json
```diff
@@ -333,7 +337,7 @@
       {
         "key": "tab",
         "command": "tabnine.tab-override",
-        "when": "tabnine.tab-override && suggestWidgetVisible && textInputFocus"
+        "when": "tabnine.tab-override && suggestWidgetHasFocusedSuggestion && suggestWidgetVisible && textInputFocus"
```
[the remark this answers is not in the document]
dimacodota: yes, noticed it while was testing my functionality, the command was triggered if the  popup was not focused

src/tryGetDebouncedCompletions.ts
```diff
@@ -0,0 +1,61 @@
...
+  const debounceMilliseconds = getDebounceValue();
+  const debounceTime = Math.max(debounceMilliseconds - time, 0);
+  return debounceTime;
+}
+
+function getDebounceValue(): number {
+  const debounceMilliseconds = vscode.workspace
+    .getConfiguration()
+    .get<number>("tabnine.debounceMilliseconds");
+  const isAlphaCapabilityEnabled = isCapabilityEnabled(
```
yairco1990, line 53: why don't you use the debounce value capabilities?

yonip23: oh that's right - for experiments ☝🏻

src/tryGetDebouncedCompletions.ts
yairco1990, line 28: Why don't you return the previously calculated value? I see the token has not cancelled yet

dimacodota: no, it is irrelevant to the offset,  it is relevant to the binary cache. 
imaging the following use-case: the first request was fired at T0 the result was an empty suggestion because the server did not complete the request,  the sever completed the request at T1 and the cash was filled with the result, if we will re-query the binary after the debounce  there is more chance that we will get the competed  response

src/utils/utils.ts
```diff
@@ -86,3 +86,18 @@ export function escapeTabStopSign(value: string): string {
 export function isMultiline(text?: string): boolean {
   return text?.includes("\n") || false;
 }
+
+export async function timed<T>(
```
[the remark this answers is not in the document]
dimacodota: no as part of the stdlib and without some 3rd party utility, `console.time()` just outputs the measured time to stdio

src/tryGetDebouncedCompletions.ts
dimacodota: bc of the re-fetching getInlineCompletionItems, I'm not sure how generic it could be

src/tryGetDebouncedCompletions.ts
dimacodota: in what case?  after debounce? or n case of cacelation ?

package.json
```diff
@@ -43,8 +43,7 @@
     "email": "support@tabnine.com"
   },
   "engines": {
-    "vscode": "^1.50.0",
-    "node": "^16.10.0"
```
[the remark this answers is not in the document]
dimacodota: it is redundant https://code.visualstudio.com/api/references/extension-manifest

dimacodota:
@yonip23 @yairco1990 - pls review after my fixes 🙏

src/test/suite/completion.driver.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+      return undefined;
+    },
+    update: sinon.fake(),
+    inspect: sinon.fake(),
+    has: sinon.fake(),
+  });
+}
+
+export function setup(): void {
```
yonip23, line 33: rename to a more specific function `setupForCompletionsTests` (same with `teardown`)

src/test/suite/completion.test.ts
```diff
@@ -377,6 +351,29 @@ describe("Should do completion", () => {
...
+      anAutocompleteResponse("d", "data"),
+      anAutocompleteResponse("da", "data1")
+    );
+
+    await openADocWith("const ", "text");
+
+    await vscode.commands.executeCommand("type", {
+      text: "d",
+    });
+    await sleep(200);
```
yonip23, line 367: extract `200` to const

src/test/suite/completion.test.ts
```diff
@@ -377,6 +351,29 @@ describe("Should do completion", () => {
...
+    await vscode.commands.executeCommand("type", {
+      text: "a",
+    });
+    await emulationUserInteraction();
+
+    verify(
+      stdinMock.write(new SimpleAutocompleteRequestMatcher("const d"), "utf8")
```
yonip23, line 374: should have been `"const da"`, no? that's the whole point of the debounce, to send only the last request

src/provideInlineCompletionItems.ts
```diff
@@ -34,11 +34,7 @@ export default async function provideInlineCompletionItems(
       return await getLookAheadSuggestion(document, completionInfo, position);
     }
 
-    const completions = await tryGetDebouncedCompletions(
-      document,
-      position,
-      token
-    );
+    const completions = await debounceCompletions(document, position, token);
```
yonip23, line 37: i see you haven't generalized the debounce function - not a blocker for this pr, just making sure you haven't forgotten it

yonip23 (approved):
lgtm, looks nice! 
had some small comments

src/test/suite/completion.test.ts
dimacodota: no, the debounce is on the "rendering" (== before the suggestion is returned to the provider) 
the binary is queried at most twice and at least once, at the following points : 
 - the completion provider is queried for competition
 - after debouncing - in order to provide the latest cached results
 if between the two calls a cancelation was issued the second request will be canceled

src/provideInlineCompletionItems.ts
dimacodota: see my comment [here](https://github.com/codota/tabnine-vscode/pull/993#discussion_r1057174733)  bc of the unique way of calculating the debounce value (debounceMs - time of current query) and that i do want to re fetch completions after the debounce, i do not think there is a real value of extracting it, it will converts this function  to less readable

src/capabilities/capabilities.ts
```diff
@@ -32,6 +32,11 @@ export enum Capability {
   SAVE_SNIPPETS = "save_snippets",
   BETA_CAPABILITY = "beta",
   FIRST_SUGGESTION_DECORATION = "first_suggestion_hint_enabled",
+  DEBOUNCE_VALUE_300 = "DEBOUNCE_VALUE_300",
```
yairco1990, line 35: I'm not sure if this is converted to lowercase somewhere, but this is how it should be:
image

PR #999, opened 2022-12-28
DEV2-1940: Add handling of open hub with action

ofekby:


src/notifications/executeNotificationAction.ts
```diff
@@ -1,23 +1,44 @@
...
-  if (selectedActions?.includes(MessageActions.OPEN_NOTIFICATIONS_IN_HUB)) {
+  if (selectedActions?.includes(MessageActionsEnum.OPEN_NOTIFICATIONS_IN_HUB)) {
     return openHub(
       StateType.NOTIFICATION,
       `/home?${NOTIFICATIONS_OPEN_QUERY_PARAM}`
     )();
   }
 
+  const openHubWithActions = selectedActions
+    ?.map((action) => action as OpenHubWithAction)
```
assapir, line 25: This seems like a bug? What if the message don't have the action? Can we actually filter by something and not only make TS happy?

assapir: Make sure it has queryParams and path?

src/notifications/executeNotificationAction.ts
```diff
@@ -1,23 +1,44 @@
...
+    .filter((action) => action.OpenHubWith)
+    .map(({ OpenHubWith: { path, query_params } }) =>
+      openHub(StateType.NOTIFICATION, buildFullHubPath(path, query_params))()
+    );
+
+  if (openHubWithActions?.length) {
+    return Promise.all(openHubWithActions).then();
```
assapir, line 32: Why the `then`?

src/notifications/executeNotificationAction.ts
```diff
@@ -1,23 +1,44 @@
...
+  }
+
   return Promise.resolve();
 }
+
+function buildFullHubPath(path: string, params: [string, string][]) {
+  return `${path}?${buildQueryParamsString(params)}`;
+}
+
+function buildQueryParamsString(params: [string, string][]) {
```
assapir, line 42: Use URLSearchParams?

assapir (approved):
If possible we should add tests

src/notifications/executeNotificationAction.ts
ofekby: To change it from returning `Promise<void[]>` to `Promise<void>`

src/notifications/executeNotificationAction.ts
ofekby: It is just typescript, once we have this weird mixed array you need to actually check if the property exists on the object and that the way to identify it.

src/notifications/executeNotificationAction.ts
ofekby: Nice

amirbilu (approved):

PR #1000, opened 2023-01-01
DEV2-1533 report suggestion about to shown

dimacodota:
DEV2-1533 report suggestion about to shown


src/aboutToRenderDecorator.ts
```diff
@@ -0,0 +1,38 @@
...
+import TabnineInlineCompletionItem from "./inlineSuggestions/tabnineInlineCompletionItem";
+import suggestionShown from "./binary/requests/suggestionShown";
+import CompletionOrigin from "./CompletionOrigin";
+
+type DecoratedFunction<R, T1 extends vscode.TextDocument, T2, T3, T4> = (
+  document: T1,
+  ...args: [T2, T3, T4]
+) => Promise<R>;
+
+export default function aboutToRenderDecorator<
```
dimacodota, line 11: this is the main functionality

src/inlineSuggestions/documentChangesTracker/DocumentTextChangeContent.ts
```diff
@@ -50,6 +50,7 @@ export default class DocumentTextChangeContent {
       !!this.contentChange &&
       // in case of /t the rangeLength will be 1, in case of spaces the rangeLength will be tabsize
       this.contentChange.rangeLength > 0 &&
+      this.contentChange.range.isSingleLine &&
```
dimacodota, line 53: this is an unrelated fix

dimacodota: ill add a test for it in a different pr

src/binary/requests/suggestionShown.ts
```diff
@@ -0,0 +1,19 @@
...
+  };
+}
+
+function suggestionShown(
+  request: SuggestionShown
+): Promise<unknown | undefined | null> {
+  return tabNineProcess.request<unknown, SuggestionShown>(request);
+}
+
+export default suggestionShown;
```
yonip23, line 19: why on a separate statement?

src/aboutToRenderDecorator.ts
yonip23, line 11: why like this? why not inline this functionality inside `provideInlineCompletionItems`?
this is much less readable

src/test/suite/completion.test.ts
```diff
@@ -392,6 +393,22 @@ describe("Should do completion", () => {
       )
     ).twice();
   });
+  it.only("should report suggestion about to shown after debounce", async () => {
```
yonip23: remove `.only`

src/test/suite/completion.test.ts
```diff
@@ -392,6 +393,22 @@ describe("Should do completion", () => {
...
+    mockGetDebounceConfig(SHORT_DEBOUNCE_VALUE);
+    mockAutocomplete(requestResponseItems, anAutocompleteResponse("d", "data"));
+    await openADocWith("const ", "text");
+    await emulationUserInteraction();
+
+    await vscode.commands.executeCommand("type", {
+      text: "d",
+    });
+    await sleep(WAIT_LONGER_THAT_DEBOUNCE);
+    await emulationUserInteraction();
```
yonip23, line 406: this function sleeps as well, why sleep twice?

src/test/suite/completion.test.ts
```diff
@@ -392,6 +393,22 @@ describe("Should do completion", () => {
       )
     ).twice();
   });
+  it("should report suggestion about to shown after debounce", async () => {
```
yonip23, line 396: please add a test that the event is not fired if the completion has been canceled (i.e only send events on completions that are actually about to be shown and not canceled)

src/test/suite/completion.test.ts
```diff
@@ -44,13 +44,15 @@ import {
   setupForCompletionsTests,
   mockGetDebounceConfig,
 } from "./completion.driver";
+import { SuggestionShownRequestMatcher } from "./utils/SuggestionShownRequestMatcher";
 
 describe("Should do completion", () => {
   const SPACES_INDENTATION = "    ";
   const TAB_INDENTATION = "\t";
-  const WAIT_LONGER_THAT_DEBOUNCE = 250;
+  const WAIT_LONGER_THAN_SHORT_DEBOUNCE = 350;
```
yonip23, line 52: remove `WAIT` from const name, its not a command, its a const (same for line 55)

src/provideInlineCompletionItems.ts
```diff
@@ -31,10 +32,17 @@ export default async function provideInlineCompletionItems(
...
+        document,
+        completionInfo,
+        position
+      );
+      reportSuggestionShown(document, result);
+      return result;
     }
 
     const completions = await debounceCompletions(document, position, token);
+    reportSuggestionShown(document, completions);
```
yonip23, line 45: are we sure the suggestion is indeed shown at this point (same for line 41)? I think vscode has its own debounce+cancel mechanism... this function just registers the plugin's suggestions to vscode - they decided later whether or not to present them, no?

yonip23 (approved):
lgtm - see my question in the comment

src/provideInlineCompletionItems.ts
dimacodota: yes, you're right, it's the closest to rendering we have right now (the shown callback will be released by the vscode team), but from what I understand from debugging their API their debounce is before this callback is executed. so, for now, it's as good as it gets

PR #1003, opened 2023-01-02
DEV2-1533 - report the same suggestion only once

dimacodota:
DEV2-1533 - report the same suggestion only once

### DIFF
--- a/src/reportSuggestionShown.ts
+++ b/src/reportSuggestionShown.ts
@@ -2,6 +2,9 @@ import * as vscode from "vscode";
 import TabnineInlineCompletionItem from "./inlineSuggestions/tabnineInlineCompletionItem";
 import suggestionShown from "./binary/requests/suggestionShown";
 import CompletionOrigin from "./CompletionOrigin";
+import { ResultEntry } from "./binary/requests/requests";
+
+let lastShownSuggestion: ResultEntry | undefined | null;
 
 export default function reportSuggestionShown(
   document: vscode.TextDocument,
@@ -9,7 +12,8 @@ export default function reportSuggestionShown(
 ): void {
   const item = completions?.items[0]?.suggestionEntry;
 
-  if (item) {
+  if (item && item.new_prefix !== lastShownSuggestion?.new_prefix) {
+    lastShownSuggestion = item;
     void suggestionShown({
       SuggestionShown: {
         origin: item.origin ?? CompletionOrigin.UNKNOWN,
@@ -18,5 +22,7 @@ export default function reportSuggestionShown(
         filename: document.fileName,
       },
     });
+  } else {
+    lastShownSuggestion = null;
   }
 }
--- a/src/reportSuggestionShown.ts
+++ b/src/reportSuggestionShown.ts
@@ -12,7 +12,7 @@ export default function reportSuggestionShown(
 ): void {
   const item = completions?.items[0]?.suggestionEntry;
 
-  if (item && item.new_prefix !== lastShownSuggestion?.new_prefix) {
+  if (item && lastShownSuggestion?.new_prefix.endsWith(item.new_prefix)) {
     void suggestionShown({
       SuggestionShown: {
         origin: item.origin ?? CompletionOrigin.UNKNOWN,
--- a/src/reportSuggestionShown.ts
+++ b/src/reportSuggestionShown.ts
@@ -13,7 +13,6 @@ export default function reportSuggestionShown(
   const item = completions?.items[0]?.suggestionEntry;
 
   if (item && item.new_prefix !== lastShownSuggestion?.new_prefix) {
-    lastShownSuggestion = item;
     void suggestionShown({
       SuggestionShown: {
         origin: item.origin ?? CompletionOrigin.UNKNOWN,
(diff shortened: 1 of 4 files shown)


src/reportSuggestionShown.ts
```diff
@@ -18,5 +22,7 @@ export default function reportSuggestionShown(
         filename: document.fileName,
       },
     });
+  } else {
+    lastShownSuggestion = null;
```
yonip23, line 26: why? if there are 3 identical suggestions you'll report the 1st and 3rd ones, because in the 2nd time you set `lastShwonSuggestion` to null, and then in the 3rd time this condition will evaluate to true: `item.new_prefix !== lastShownSuggestion?.new_prefix`

src/test/suite/completion.test.ts
```diff
@@ -438,6 +438,26 @@ describe("Should do completion", () => {
       stdinMock.write(new SuggestionShownRequestMatcher("dom"), "utf8")
     ).once();
   });
+  it("should report suggestion shown only once for the same suggestion", async () => {
```
yonip23, line 441: im starting to get the feeling that we should split this file, lets move all debounce related tests to a separate test file?

yonip23 (commented):
see comments

src/reportSuggestionShown.ts
yairco1990: @yonip23 is right. maybe you meant to set it to null just in case there is no item (without checking the prefixes)

dimacodota: you're right, I wanted to init it in any case with the latest suggestion, modified the test to test this

src/test/suite/completion.test.ts
dimacodota: will be handled in a different pr - https://tabnine.atlassian.net/browse/DEV2-2078

src/test/suite/completion.test.ts
```diff
@@ -438,6 +438,31 @@ describe("Should do completion", () => {
...
+    mockAutocomplete(
```
yairco1990, line 442: I'm not sure, but isn't the response is:
```
      anAutocompleteResponse("blibliblid", "ata"),
      anAutocompleteResponse("blibliblida", "ta"),
      anAutocompleteResponse("blibliblidat", "a")
```
?

yonip23: @yairco1990 you're right and wrong - as long as its the same word, the `new_prefix` will remain the same, and the `old_suffix` will change.
But if you type beyond the first word, the `new_prefix` will change as well.
@dimacodota wdyt? this means that you'll still fire the shown even twice for the same logical completion

yonip23 (approved):
lgtm, see @yairco1990 's thread

src/test/suite/completion.test.ts
yairco1990: if we are not rendering a new completion in this case it is an issue

dimacodota: not exactly,  the binary will return the common text between the current text and the suggestion as `old_prefix` that should be replaced: see an example: 
```
"const da"  -> {"old_prefix":"da","results":[{"new_prefix":"data =" ...}]
"const dat"  -> {"old_prefix":"dat","results":[{"new_prefix":"data ="...}]
```

but there are places(especially after whitespaces ` `) that although it is the same suggestion, it will not contain the  `old_prefix`  like in the following example:
```
"const data" -> {"old_prefix":"data","results":[{"new_prefix":"data = new"..}]
"const data " -> {"old_prefix":"","results":[{"new_prefix":"= new"..}]
```
added a fix for the second case(ie the prev suggestion end with the current)

dimacodota: handled here https://github.com/codota/tabnine-vscode/pull/1003/commits/1fdeb81cdb4a6e1660b5430906ecac327be350dc

yairco1990: great 👍
and what happens if the suffix has changed? doesn't it mean that we just rendered a new completion but we don't fire the event?

dimacodota: Of course im missing the "not" there

dimacodota: @yairco1990 - can you please elaborate on what you've meant?   the event should be fired in case the prev suggestion is not ending with the current suggestion

yonip23 (approved):
Maybe add tests for the "endswith" condition? 
1. First time -> "abc", second time -> "bc", assert that only one event is fired.
2. First time -> "abc", second time -> "bcd", assert that 2 events fired

Not critical though, lgtm

src/test/suite/completion.test.ts
yairco1990: what I meant is that you recognize there is no change in the suggestion based only on the prefix. my question is if there is an option this condition is evaluated to "true", but the suffix has changed, and therefore, it should be counted as a new suggestion.

dimacodota: oh you mean if the `new_suffix` will be different?

yairco1990: yep

dimacodota: oh, the `new_suffix` handles the closing brackets in single-line suggestions,  not sure how common is the case of the same suggestion w/o closing brackets. so ill merge it as is

PR #1016, opened 2023-01-23
DEV2-2022 use new api version 4.4.223

yonip23:
exactly the same changes as here: https://github.com/codota/tabnine-intellij/pull/462


yonip23:
I checked that if I work with binary API version 4.4.71 I get basic suggestions without crashing (this is the last commit ^)
before my fix, we crashed on NPE.
@yairco1990 @dimacodota this ensures that even if we're in the new plugin old binary situation at least the user will get suggestions.

I haven't seen any other broken behavior in the plugin, but can't guarantee its safe either.

yairco1990 (approved):
lgtm

dimacodota (approved):
lgtm

PR #1017, opened 2023-01-29
DEV2-2244 sandboxed version of the app

avnerbarr:


src/cloudEnvs/openLogin.ts
```diff
@@ -3,8 +3,12 @@ import { URL } from "url";
 import { StateType, TABNINE_URL_QUERY_PARAM } from "../globals/consts";
 import createHubWebView from "../hub/createHubWebView";
 import hubUri from "../hub/hubUri";
+import {isSandboxed} from "../sandbox";
 
 export default async function openLogin(): Promise<void> {
+  if (isSandboxed()) {
```
yanir-codota, line 9: Is this enough? 
I think we want to remove the option altogether

src/sandbox.ts
```diff
@@ -0,0 +1,11 @@
+export function isSandboxed(): boolean {
+  // need to figure out how to inject this during the packaging phase.
+  // see feature flags here: // https://webpack.js.org/plugins/define-plugin/#feature-flags
+  // vsce isn't passing the argument to the vscode:prepublish phase (see package.json)
+  // see this question: https://github.com/microsoft/vscode-vsce/issues/818
+  // consider this post: https://moriz.xyz/blog/feature-flags-in-nextjs
+  if (SANDBOX) {
```
amirbilu, line 7: this can be webpack build argument

amirbilu: Okay.. I saw this is the past you've taken

yanir-codota:
What about loading from disk (bundles)?

src/cloudEnvs/openLogin.ts
avnerbarr: If you mean in respect to a user interaction then yes. but there are internal code paths which lead to hear and this will block that

src/sandbox.ts
avnerbarr: I'm not sure how to make this work in webpack and with vsce . do you know how? You can reference the comments here

avnerbarr:
> What about loading from disk (bundles)?

I didn't change any of the loading paths. I am assuming that there will be some other script which will copy the relevant binaries to the proper paths. Also I will be adding the flag that bilu created

PR #1030, opened 2023-02-19
Avner/DEV2-2244 onprem

avnerbarr:


.github/workflows/packageOnPrem.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.19'
+      #          runs the binaries downloader script which will be packaged into this plugin
+      - name: pull binaries
+        run: VERSION=${{ inputs.binaryVersion }} ./binaries/downloadBinaries.sh
+      #      https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
+      - name: modify readme
+        run: |
+          echo "" > ./README.md
```
assapir, line 30: Why deleting the current Readme and not just adding to it?

buildOnPrem.sh
```diff
@@ -0,0 +1,3 @@
+yarn onprem:prep \
+&& yes | vsce package \
```
assapir, line 2: Make this file a yarn script?

package.json
```diff
@@ -307,6 +320,10 @@
             "default": null,
             "description": "Experimental: Enable modules auto-import following applying Tabnine suggestions"
           },
+          "tabnine.cloudHost": {
+            "type": "string",
+            "description": "The tabnine cloud host to connect to"
```
assapir, line 325: Can we expose that only if on prem?

src/binary/runBinary.ts
```diff
@@ -13,14 +14,17 @@ export default async function runBinary(
...
   const context = getTabnineExtensionContext();
   const args: string[] = [
-    "--client=vscode",
+    ON_PREM ? "--client=vscode_on_prem" : "--client=vscode",
     "--no-lsp=true",
     tabnineExtensionProperties.logFilePath
       ? `--log-file-path=${tabnineExtensionProperties.logFilePath}`
       : null,
+    ON_PREM ? "--onprem=true" : null,
+    ON_PREM ? "--no_bootstrap" : null,
```
assapir, line 23: No need to pass it, it won't bootstrap anyway

src/binary/runBinary.ts
```diff
@@ -13,14 +14,17 @@ export default async function runBinary(
...
     tabnineExtensionProperties.logLevel
       ? `--log-level=${tabnineExtensionProperties.logLevel}`
       : null,
+    tabnineExtensionProperties.cloudHost ? `--cloud2_url=${tabnineExtensionProperties.cloudHost}` : null,
```
assapir, line 27: Pass --onprem only if that set?

src/onPrem.ts
```diff
@@ -0,0 +1,2 @@
+var PREM = require('./onPremJ.js');
```
assapir, line 1: Wait what?
Also, use const, not var

.github/workflows/packageOnPrem.yml
avnerbarr: because it has lots of links and junk

buildOnPrem.sh
avnerbarr: can't because it is modifying the package.json and that didn't work

package.json
avnerbarr: yes it can be possible to add in the prep phase

src/binary/runBinary.ts
avnerbarr: I'm not sure about that, depends on the build

src/binary/runBinary.ts
avnerbarr: the other way around otherwise it won't work in regular mode. fixed that

src/onPrem.ts
avnerbarr: will that work with a require call?

package.json
avnerbarr: any idea how to append a nested object to this json from a cli call? (yarn,npm etc.) (without installing some fancy tool I suppose)

avnerbarr: I suppose this might work. I'll try 
https://www.npmjs.com/package/transform-json-webpack-plugin

avnerbarr:
moved to this PR because of JIRA rule
https://github.com/codota/tabnine-vscode/pull/1033

PR #1033, opened 2023-02-20
DEV2-2244 onprem

avnerbarr:


.eslintignore
```diff
@@ -5,4 +5,6 @@ dist
 out
 # don't lint nyc coverage output
 coverage
-.eslintrc.js
\ No newline at end of file
+.eslintrc.js
+src/onPremJ.js
```
dimacodota, line 9: why?

.github/workflows/packageOnPrem.yml
```diff
@@ -0,0 +1,50 @@
...
+    name: Make vsix
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.19'
+      #          runs the binaries downloader script which will be packaged into this plugin
+      - name: pull binaries
+        run: VERSION=${{ inputs.binaryVersion }} ./binaries/downloadBinaries.sh
```
dimacodota, line 26: sync with @assapir - the on prem binary is not released to ga, i think you should download it from the acpl release attachments

package.json
```diff
@@ -96,7 +96,10 @@
   "main": "./out/extension",
   "scripts": {
     "analyze:bundle": "webpack --env analyzeBundle",
-    "vscode:prepublish": "yarn --frozen-lockfile && yarn clear-out && webpack --mode production",
+    "vscode:prepublish": "yarn --frozen-lockfile && yarn clear-out && yarn webpack",
+    "webpack": "yarn webpackGA",
+    "webpackOnPrem": "webpack --mode production --env onPrem",
```
dimacodota, line 101: `webpackOnPrem`-> `webpack:OnPrem`

package.json
```diff
@@ -96,7 +96,10 @@
...
+    "webpackGA": "webpack --mode production",
```
dimacodota, line 102: `webpackGA` -> `webpack:GA`

package.json
```diff
@@ -107,6 +110,16 @@
     "test:prepare": "yarn clear-out && tsc && yarn test:copyassets",
     "test": "yarn test:prepare && node ./out/test/runTest.js",
     "vsce:package": "vsce package",
+    "onprem:prep": "yarn onprem:prep:package && yarn onprem:prep:readme",
```
dimacodota, line 113: would suggest to do all of it in a `buildOnPrem.js` file because it is not readable, this file shold be not coompiled

webpack.config.js
```diff
@@ -63,6 +63,18 @@ const config = {
 };
 
 module.exports = (env) => {
+
+  if (env.onPrem) {
+    console.log("building for onprem");
+    config.entry.extension = "./src/extensionOnPrem.ts";
+    config.plugins.push(new webpack.DefinePlugin({
+      __ONPREM__: JSON.stringify(true)
```
dimacodota, line 71: why here and not in the action ?

src/binary/paths.ts
```diff
@@ -5,68 +5,103 @@ import {
...
 const BUNDLE_SUFFIX = getBundleSuffix();
 
 export async function setBinaryRootPath(
   extensionContext: vscode.ExtensionContext
 ): Promise<void> {
-  binaryRootPath =
+  if (ONPREM) {
+    const base = `binaries/${getArch()}-${getPlatform()}`;
+    BINARY_ROOT_PATH = path.join(extensionContext.extensionPath, base);
+    await makeExecutable(BINARY_ROOT_PATH);
```
dimacodota, line 21: already implemented use  this https://github.com/codota/tabnine-vscode/blob/dc35e8a59c7ecd0e074c893fc0a58f5bfb3e1c20/src/binary/binaryFetcher/bundleDownloader.ts#L88

src/binary/paths.ts
```diff
@@ -5,68 +5,103 @@ import {
...
+  }
+  const files = await fs.readdir(bundleDirectory);
+  console.log("making files executable", files);
+  await Promise.all(
+    files.map((file) => fs.chmod(path.join(bundleDirectory, file), 0o755))
+  );
+}
+
+export function bundledTabnineBinaryPath(): string {
+  return path.join(<string>BINARY_ROOT_PATH, binaryName());
```
dimacodota, line 49: BINARY_ROOT_PATH can be undefined

src/binary/paths.ts
```diff
@@ -5,68 +5,103 @@ import {
...
 const ARCHITECTURE = getArch();
 const SUFFIX = getSuffix();
 const BUNDLE_SUFFIX = getBundleSuffix();
 
 export async function setBinaryRootPath(
   extensionContext: vscode.ExtensionContext
 ): Promise<void> {
-  binaryRootPath =
+  if (ONPREM) {
+    const base = `binaries/${getArch()}-${getPlatform()}`;
```
dimacodota, line 19: why ? all this is already implemented, you can use this https://github.com/codota/tabnine-vscode/blob/dc35e8a59c7ecd0e074c893fc0a58f5bfb3e1c20/src/binary/paths.ts#L29

src/binary/runBinary.ts
```diff
@@ -13,14 +14,19 @@ export default async function runBinary(
 
   const context = getTabnineExtensionContext();
   const args: string[] = [
-    "--client=vscode",
+    ONPREM ? "--client=vscode_on_prem" : "--client=vscode",
     "--no-lsp=true",
     tabnineExtensionProperties.logFilePath
       ? `--log-file-path=${tabnineExtensionProperties.logFilePath}`
       : null,
+    ONPREM ? "--onprem=true" : null,
```
dimacodota, line 22: i think it is irelevant

src/binary/runBinary.ts
```diff
@@ -13,14 +14,19 @@ export default async function runBinary(
...
+    ONPREM ? "--no_bootstrap" : null,
```
dimacodota, line 23: i think it is irelevant

src/binary/runProcess.ts
```diff
@@ -22,6 +22,12 @@ export function runProcess(
     return require("./mockedRunProcess").default() as BinaryProcessRun;
   }
   report(EventName.START_BINARY);
+  console.log(
```
dimacodota, line 25: remove

src/extensionOnPrem.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function initStartup(context: vscode.ExtensionContext): void {
+  setTabnineExtensionContext(context);
+}
+
+async function backgroundInit(context: vscode.ExtensionContext) {
+  await setBinaryRootPath(context);
+  await initBinary();
+  notifyBinaryAboutWorkspaceChange();
```
dimacodota, line 35: this is irelevant for ton prem, it is being used in the right side panel

src/notifications/pollNotifications.ts
```diff
@@ -32,6 +33,9 @@ export function cancelNotificationsPolling(): void {
 export async function doPollNotifications(
   context: vscode.ExtensionContext
 ): Promise<void> {
+  if (ONPREM) {
```
dimacodota, line 36: irelevant you do not call pollNotifications

src/commandsHandler.ts
```diff
@@ -13,13 +14,14 @@ export const SAVE_SNIPPET_COMMAND = "Tabnine.saveSnippet";
...
   context: ExtensionContext
 ): Promise<void> {
-  context.subscriptions.push(
-    commands.registerCommand(CONFIG_COMMAND, openHub(StateType.PALLETTE))
-  );
-
-  context.subscriptions.push(
-    commands.registerCommand(STATUS_BAR_COMMAND, handleStatusBar(context))
-  );
+  if (!ONPREM) {
```
dimacodota, line 17: remove it from here and just do not call `registerCommands` from the extensionOnPrem.ts

src/reports/reporter.ts
```diff
@@ -26,6 +27,9 @@ export function initReporter(
   if (inTestMode) {
     return;
   }
+  if (ONPREM) {
```
dimacodota, line 30: change the `reporter` to be `TelemetryReporter | undefined` and then check if it is not `undefined` instead of duplicate the check

src/statusBar/StatusBarData.ts
```diff
@@ -56,6 +58,16 @@ export default class StatusBarData {
   }
 
   private updateStatusBar() {
+    if (ONPREM) {
+      const issueText = this._text ? `: ${this._text}` : "";
+      const limited = this._limited ? ` ${LIMITATION_SYMBOL}` : "";
```
dimacodota, line 63: _limited is irelevant

src/extensionOnPrem.ts
```diff
@@ -0,0 +1,52 @@
...
+import installAutocomplete from "./autocompleteInstaller";
+import { handleSelection, notifyBinaryAboutWorkspaceChange } from "./extension";
+
+export async function activate(
+  context: vscode.ExtensionContext
+): Promise<void> {
+  console.log("*****THIS IS AN ONPREM BUILD****");
+  void initStartup(context);
+  handleSelection(context);
+  registerStatusBar(context);
```
dimacodota, line 21: i would override this function and and create an onprem `registerStatusBar` function instead of modifying the `updateStatus` function

webpack.config.js
```diff
@@ -63,6 +63,18 @@ const config = {
 };
 
 module.exports = (env) => {
+
+  if (env.onPrem) {
+    console.log("building for onprem");
+    config.entry.extension = "./src/extensionOnPrem.ts";
```
dimacodota, line 69: this is an uncommon practice to do it , you can just modify the https://github.com/codota/tabnine-vscode/blob/dc35e8a59c7ecd0e074c893fc0a58f5bfb3e1c20/webpack.config.js#L14
and do something like ``"./src/extension${env.onPrem ? "OnPrem" : ""}.ts"``

src/reports/reporter.ts
avnerbarr: then you would be led to believe that it should be defined. this check makes the intention obvious and explicit

src/commandsHandler.ts
avnerbarr: some of the commands need to be registered

src/notifications/pollNotifications.ts
avnerbarr: it is called on every selection suggestion

src/extensionOnPrem.ts
avnerbarr: I'll remove. it isn't aware what the point of this class is

src/binary/runBinary.ts
avnerbarr: we are still passing the flags. if it gets changed I will change here

src/binary/runBinary.ts
avnerbarr: not yet

src/binary/paths.ts
avnerbarr: there is no version . it is bundled flat

src/binary/paths.ts
avnerbarr: this is the current convention. if it could happen now, it could happen before. the design isn't great to have dangling references

dimacodota:
General notes: 
 - sync with @assapir  on the onprem binary build location
 - since you already created a different entry point for the onprem extension, idd try to manage the "onprem" decisions on this level and not  on the functional level where you  need to add the `if(onprem)` everywhere

webpack.config.js
avnerbarr: GitHub action? How could it be in the action, when the web pack is doing the replacements

package.json
avnerbarr: because there are shell commands here and this mechanism allows for running from outside individually

.eslintignore
avnerbarr: irrelevant now. will remove it

src/binary/runProcess.ts
avnerbarr: what's the problem with this log

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "publisher": "TabNine",
   "name": "tabnine-vscode",
-  "version": "3.6.38",
+  "version": "3.6.40",
```
dimacodota, line 4: revert it and use the bump action to release

src/binary/paths.ts
dimacodota: ok , but all the path logic is already implemented there, you can extract it or something

src/binary/paths.ts
dimacodota: i meant the casting you are doing here, it will be more clear to adjust the return type and this casting

src/binary/paths.ts
```diff
@@ -5,68 +5,92 @@ import {
...
   } catch (err) {
     // Exception is thrown if the path already exists, so ignore error.
   }
 }
 
+export function bundledTabnineBinaryPath(): string {
+  return path.join(<string>BINARY_ROOT_PATH, binaryName());
+}
+
+function binaryName(): string {
```
dimacodota, line 41: getSuffix

src/binary/binaryFetcher/existingVersionHandler.ts
```diff
@@ -1,12 +1,20 @@
...
 import { getRootPath, versionPath } from "../paths";
+import { setDirectoryFilesAsExecutable } from "../utils";
 
 export default async function handleExistingVersion(): Promise<string | null> {
   try {
     const versionPaths = await fs.readdir(getRootPath());
+    await Promise.all(
+      versionPaths
+        .map((version) => path.dirname(versionPath(version)))
+        .filter((p) => existsSync(p))
```
dimacodota, line 15: why sync?

src/binary/binaryFetcher/existingVersionHandler.ts
```diff
@@ -1,12 +1,20 @@
...
+        .map(async (p2) => setDirectoryFilesAsExecutable(p2))
```
dimacodota, line 16: change to `map(setDirectoryFilesAsExecutable)`

src/commandsHandler.ts
dimacodota: what commands? enableSaveSnippetContext is not releavant

src/extensionOnPrem.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+async function backgroundInit(context: vscode.ExtensionContext) {
+  await setBinaryRootPath(context);
+  await initBinary();
+
+  setDefaultStatus();
+  void registerCommands(context);
+  await installAutocomplete(context);
+  vscode.languages.registerHoverProvider(
```
dimacodota, line 39: this not relevant to onprem

src/binary/binaryFetcher/existingVersionHandler.ts
dimacodota: use https://github.com/codota/tabnine-vscode/blob/8e1cdc07931034dc78c552327c2b6ccee2e12e53/src/utils/file.utils.ts#L28

src/binary/binaryFetcher/existingVersionHandler.ts
dimacodota: must say it is a bit strange to manage it here and the downloading in a different part of the code, id suggest to to make the files  executable after  you bundle them, somewhere in the ci. and keep the  downloading/bundling/chmod together because they are related

avnerbarr: Not possible. Chmod is platform dependent and that is why the downloader does it as well. The chmod has to occur somewhere and this is the first place it is possible. I have to admit that the "duplicated" code was simpler by 10x and that is why i did it. the amount of changes required in order to not "duplicate" the path.join trivial code are immense in comparison  and not worth the "smell" of the dup. 
In other words IMO the cost of "duplicating" trivial code of 1 line vs changes that need to be propagated to tens of places are definitely not worth it

dimacodota: 1 -the downloader does the chmod bc the download happens after the installation.
2 - it is posible that we we did in the "bundled version, you can see it here https://github.com/codota/TabNine/blob/8ba545f20dac8de3b0fc0c6acf377a29e89bbcb4/dl_binaries.sh#LL23C30-L23C30

3 - disagree that this is a trivial code . it is is not bc you need to follow why it is here

dimacodota (approved):
lgtm, although I don't fully agree with some parts here. I do understand the time frame and I think it will be easier to fix after that

src/binary/binaryFetcher/existingVersionHandler.ts
dimacodota: btw, the duplication was not in `path.join` it was in the calculation of the arch and the platform , which  just not makes any sense

dimacodota:
you do need to modify the plugin id in the package.json

webpack.config.js
avnerbarr: there is no env object there

PR #1035, opened 2023-02-21
DEV2-2328 - code gen fixes

dimacodota:
DEV2-2328 - code gen fixes

this ps contains the following fixes:
 - do not re create the authentication session on each request
 - in case of class methods the payload will hold the class as the `text` and the method as `block`
 - in case of functions the `text` filed will be empty


yahave (approved):
lgtm

PR #1040, opened 2023-02-22
fix

avnerbarr:


binaries/downloadBinaries.sh
```diff
@@ -8,14 +8,16 @@ echo "script dir is $SCRIPT_DIR"
...
 do
-    cd $SCRIPT_DIR/${target} \
-            && unzip TabNine.zip \
-            && chmod +x * \
-            && rm TabNine.zip \
-            && rm *local* \
-            && cd -
+    mkdir -p $SCRIPT_DIR/$VERSION/${target} \
+    && cd $SCRIPT_DIR/$VERSION/${target} \
+    && gsutil -m cp -r "gs://latest-onprem-binaries/$VERSION/latest/${target}/TabNine.zip" . \
```
assapir, line 17: That... Waste of time? gsutil downloading it in parallel. I think it's just easier to move the files

avnerbarr: It is so non critical and this is how i found the bug you had. I am not near computer. If youd like to modify it feel free and test your chamges and go ahead amd merge

nircodota (approved):

PR #1054, opened 2023-03-08
DEV2-2244 show connection issue when cloud health is not OK

avnerbarr:


src/statusBar/statusBar.ts
```diff
@@ -44,11 +45,13 @@ export async function pollServiceLevel(): Promise<void> {
   if (!statusBarData) {
     return;
   }
+
+
+  const state = await getState();
   if (ONPREM) {
+    statusBarData.text = state?.cloud_connection_health_status !== "Ok" ? "Cloud connection issue" : null;
```
avnerbarr, line 52: @yonip23 @yanir-codota we need to sync on the exact text we want here.

assumption is we only are dealing with the cloud connection errors. in future if this changes we can augment this, for now seems simple and solves the issue at hand

yanir-codota, line 52: null data means standard active `tabnine enterprise` notion?

avnerbarr:
YesOn Mar 8, 2023, at 12:59 PM, yanir-codota ***@***.***> wrote:﻿
@yanir-codota commented on this pull request.



In src/statusBar/statusBar.ts:
>    if (ONPREM) {
+    statusBarData.text = state?.cloud_connection_health_status !== "Ok" ? "Cloud connection issue" : null;

null data means standard active tabnine enterprise notion?

src/statusBar/statusBar.ts
yanir-codota, line 52:
```suggestion
    statusBarData.text = state?.cloud_connection_health_status !== "Ok" ? "Server connectivity issue" : null;
```

PR #1066, opened 2023-03-19
support no proxy

amirbilu:


yanir-codota (approved):
LGTM

it would be good to have an error indication in case of invalid url. but not sure it's related to this scope.

PR #1096, opened 2023-05-02
DEV2-2653 - remove unused features

dimacodota:
- old inline api
 - save snippets (unused)
 - first inline decoration(unused)
 - welcome page (unused)


amircodota (approved):
LGTM

PR #1115, opened 2023-05-23
DEV2-2532 enterprise status bar flow

dimacodota:
This pr introduces five steps that are represented in the status bar:

1. The user needs to set the server URL. - The status bar will show a warning (yellow) indication and provide a "Set Settings" action.
2. The user needs to be logged in. - The status bar will show a warning indication and provide a login action.
3. Waiting for the binary to start. - The status bar will display a loading animation.
4. In case of any failure.  The status bar will show an error indication (red).
5. After a successful binary start, the status bar will display the label "Tabnine".

TBD - test, WIP in diferent pr

### DIFF
--- a/src/enterprise/statusBar/StatusBar.ts
+++ b/src/enterprise/statusBar/StatusBar.ts
@@ -85,11 +85,20 @@ export class StatusBar implements Disposable {
     this.item.setDefault();
     this.item.setCommand(StatusState.Ready);
     this.statusPollingInterval = setInterval(() => {
-      void getState().then((state) => {
-        if (state?.cloud_connection_health_status !== "Ok") {
-          this.item.setWarning("Server connectivity issue");
+      void getState().then(
+        (state) => {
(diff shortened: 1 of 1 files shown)


assapir (commented):
I added a UserInfo request as we spoke
https://github.com/codota/acpl-paid/pull/2702

yanir-codota:
@dimacodota can we add another error where the user is logged in but not part of the team as "you need to log in" too error?

dimacodota:
> @dimacodota can we add another error where the user is logged in but not part of the team as "you need to log in" too error?

sure ill add it

avnerbarr:
@dimacodota need a warning status bar if the user has tabnine installed along with the self-hosted version

avnerbarr:
@dimacodota is this being merged?

Can you patch this :

https://github.com/codota/tabnine-vscode/blob/206ecdbee67a365ff1e81a6a7bf416568a793144/src/enterprise/extension.ts#L140

PR #1146, opened 2023-06-15
Fix/avner fixes

dimacodota:


avnerbarr (commented):
I don't understand what is going on here

src/extension.ts
```diff
@@ -69,6 +71,9 @@ export async function activate(
   if (context.extensionMode !== vscode.ExtensionMode.Test) {
     handlePluginInstalled(context);
   }
+  const statePoller = new StatePoller();
```
avnerbarr, line 74: state poller can be global to all objects polling state. that was the point of having the statics

avnerbarr: if going this route you need to pass the poller to all the relevant places when that will be implemented (I didn't touch any of the existing pollers because of time constraints)

src/registration/forceRegistration.ts
```diff
@@ -0,0 +1,113 @@
...
+
+if (isForceEnabled.value === undefined) {
+  const disposable = onDidRefreshCapabilities(() => {
+    disposable.dispose();
+    // now take the real value since it got retrieved
+    isForceEnabled.value = isCapabilityEnabled(Capability.FORCE_REGISTRATION);
+  });
+}
+
+export function forceRegistrationIfNeeded(poller: StatePoller) {
```
avnerbarr, line 37: you need to pass the other listeners in as well if going this route

src/registration/forceRegistration.ts
```diff
@@ -0,0 +1,113 @@
...
+        disposable?.dispose();
+        if (change.currentState?.is_logged_in === false) {
+          void notifyState();
+        }
+      });
+    }
+  }
+}
+
+async function popupState() {
```
avnerbarr, line 89: you aren't aligned on the PR

src/state/statePoller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { BINARY_STATE_POLLING_INTERVAL_MILLISECONDS } from "../globals/consts";
+import { getState, tabNineProcess } from "../binary/requests/requests";
+
+export type StateEmitterProps = {
+  currentState: State | null | undefined;
+  previousState: State | null | undefined;
+};
+
+export class StatePoller implements Disposable {
```
avnerbarr, line 12: should be a singleton to avoid all the code locations doing their own polling. The polling interval should be the LCD of all of them

src/statusBar/StatusBarData.ts
```diff
@@ -71,6 +77,14 @@ export default class StatusBarData implements Disposable {
     const serviceLevel = this.getDisplayServiceLevel();
     const limited = this._limited ? ` ${LIMITATION_SYMBOL}` : "";
     this._statusBarItem.text = `${FULL_BRAND_REPRESENTATION}${serviceLevel}${this.getIconText()}${issueText.trimEnd()}${limited}`;
+    if (shouldBlockCompletions()) {
```
avnerbarr, line 80: not aligned with latest

PR #1154, opened 2023-06-19
DEV2-2899: move the chat-app to external repo

yairco1990:


dimacodota (approved):
lgtm, had one small comment regarding the usage of fetch vs.  axios

src/tabnineChatWidget/ChatViewProvider.ts
```diff
@@ -136,61 +132,10 @@ export default class ChatViewProvider implements WebviewViewProvider {
...
-        <div id="root"></div>
-      </body>
-      </html>
-      `;
+  axios.get("http://localhost:3000/index.html")
+    .then((response) => {
+      // eslint-disable-next-line no-param-reassign
+      webviewView.webview.html = response.data;
+    })
+    .catch(console.error);
```
dimacodota, line 140: ill add here a "check that you are running the chat app"  notification or something, just for smoother dev process

.github/workflows/alpha_release.yml
```diff
@@ -36,6 +36,20 @@ jobs:
       with: 
         files: ./src/consts.ts
         replacements: '<INSTRUMENTATION_KEY>=${{ secrets.INSTRUMENTATION_KEY }}'
+    - name: Get Chat Version
```
dimacodota, line 39: for a later stage, but i think you can define it as a composite action and reuse it 
https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

src/tabnineChatWidget/ChatViewProvider.ts
yairco1990: great idea
added

aarondill:
From what I can tell, the new repo under `codota/tabnine-chat-app` is currently private. Is there a particular reason for this change, when the assets were previously publicly available under this repo?

PR #1171, opened 2023-06-27
DEV2-3023 use user avatar

dimacodota:


src/tabnineChatWidget/ChatApi.ts
```diff
@@ -66,6 +67,7 @@ export function initChatApi(context: vscode.ExtensionContext) {
       return {
         token: state.access_token,
         username: state.user_name,
+        user_avatar_url: state.user_avatar_url,
```
yonip23, line 70: Use camel case when sending to the app

yonip23 (approved):

PR #1174, opened 2023-06-29
DEV2-3003 align status bar notifications

avnerbarr:


yanir-codota (commented):
LGTM. 
would love if @dimacodota would take a look too.

src/statusBar/pollStatusBar.ts
```diff
@@ -3,21 +3,25 @@ import { getStatus } from "../binary/requests/statusBar";
...
   disposeStatusBarCommand,
 } from "./statusBarActionHandler";
+import { statePoller } from "../state/statePoller";
 
 let statusPollingInterval: NodeJS.Timeout | null = null;
 
 export default function pollStatuses(
   context: vscode.ExtensionContext
 ): vscode.Disposable {
+  const statePollerDisposable = statePoller.event((change) => {
```
dimacodota, line 19: 1 - what problem does this solves?
2 - why do we need the both pollings here  ? ie statePoller and doPollStatus ?

src/registration/forceRegistration.ts
```diff
@@ -122,22 +124,35 @@ async function notifyState() {
   }
 }
 
+let awaiting = false;
 function awaitLoginNotification() {
+  if (awaiting) {
```
dimacodota, line 129: remove all the awaiting, and just return a promise which is resolved after the notifications is shown, promise is resolved only once, but you need to keep the same instance of it

src/statusBar/pollStatusBar.ts
avnerbarr: they are emitting at different intervals

src/registration/forceRegistration.ts
avnerbarr: why is that easiser / better?

src/statusBar/pollStatusBar.ts
avnerbarr: state poller is state of user

status poller is status bar messages from the binary

src/registration/forceRegistration.ts
dimacodota: it is a more pragmatic way  of managing an async flow, as we did here :
https://github.com/codota/tabnine-vscode/blob/168deb7698863341024bd49322cc2ff915f8341c/src/capabilities/capabilities.ts#L62

avnerbarr: what is the syntax for that?

src/statusBar/pollStatusBar.ts
dimacodota: ok

src/registration/forceRegistration.ts
avnerbarr: this is just a completion handler. Where is it being resolved between 2 states

avnerbarr: @dimacodota

yanir-codota (approved):

PR #1177, opened 2023-06-29
DEV2-3026: Copy configuration from updater extension

assapir:
From the code (and testing), it seems that the user doesn’t actually need to set it twice because it is looking at the other value as well. But if the user clicks on the settings, the cloud URL seems empty. So we will copy the settings if not set (so we won't override it)


ofekby (approved):

dimacodota (approved):
lgtm

PR #1188, opened 2023-07-04
DEV2-3003: Change force registration flow

ofekby:


src/statusBar/StatusBarData.ts
```diff
@@ -105,6 +109,10 @@ export default class StatusBarData implements Disposable {
   }
 
   private getDisplayServiceLevel(): string {
+    if (!isCapabilitiesReady()) {
```
assapir, line 112: we should display something.... but what happen if the user has no connection to tabnine so it will never be ready?

assapir (approved):
minor concern

src/statusBar/StatusBarData.ts
ofekby:
```
!!capabilities &&
    (!capabilities.experiment_source ||
      capabilities.experiment_source === ExperimentSource.API ||
      capabilities.experiment_source === ExperimentSource.APIErrorResponse)
```

The user must fall to one of these state very early when the binary is initialized. Null is backward compatibility with older binaries, API meaning success,  APIErrorResponse meaning the request failed for some reason (Status code, connectivity, etc...)

yanir-codota (approved):

src/statusBar/StatusBarData.ts
yanir-codota: What would be the outcome of failure here? 
what is the expected user experience? 
what will happen in a air-gapped env?

ofekby: What would be the outcome of failure here? it is a boolean value, so if true will keep the indicator with no suffix.
what is the expected user experience? While the features state is not resolved to a stable state it will keep the indicator with no suffix, meaning we can only more from "tabnine" -> "tabnine starter" or "tabnine" -> "tabnine" or "tabnine" -> "tabnine pro" or "tabnine" -> "tabnine enterprise".

what will happen in a air-gapped env? nothing changes, it is controller by separate part of the code.

PR #1192, opened 2023-07-06
DEV2-3070 add `resolve_workspace_commands` chat command

yonip23:
### Changes
1. A new request is introduced: `resolve_workspace_commands` - currently, only `symbolSearch` is supported:
```typescript
export type ResolveWorkspaceCommandsRequest = {
  workspaceCommands: WorkspaceCommandInstruction[];
};

export type WorkspaceCommandInstruction = {
  command: WorkspaceCommand;
  arg: string;
};

type WorkspaceCommand = "symbolSearch";
```
This request returns a `WorkspaceData | undefined` - currently containing only the `findSymbols` results: 
```typescript
export type WorkspaceData = {
  symbols?: string[];
};
```
2. I introduced a generic mechanism that resolves and executes these instructions.

### DIFF
--- a/src/tabnineChatWidget/handlers/resolveSymbols.ts
+++ b/src/tabnineChatWidget/handlers/resolveSymbols.ts
@@ -39,7 +39,7 @@ export async function resolveSymbols({
 }
 
 function isProbablyNotSource(path: string): boolean {
-  return ["node_modules", "dist", "build", "target", "out"].some(
-    (dir) => path.startsWith(dir) || path.startsWith(`/${dir}`)
+  return ["node_modules", "dist", "build", "target", "out"].some((dir) =>
+    path.startsWith(dir)
   );
 }
--- a/src/tabnineChatWidget/handlers/resolveSymbols.ts
+++ b/src/tabnineChatWidget/handlers/resolveSymbols.ts
@@ -1,18 +1,45 @@
 import * as vscode from "vscode";
 
+export type SymbolInformationResult = vscode.SymbolInformation & {
+  relativePath: string;
+};
+
 export async function resolveSymbols({
   symbol,
 }: {
   symbol: string;
-}): Promise<vscode.SymbolInformation[] | undefined> {
+}): Promise<SymbolInformationResult[] | undefined> {
   const symbols = await vscode.commands.executeCommand<
     vscode.SymbolInformation[]
   >("vscode.executeWorkspaceSymbolProvider", symbol);
-  return symbols?.filter((workspaceSymbol) => {
-    const workspacePath = vscode.workspace.workspaceFolders?.[0].uri.path;
-    return (
-      workspacePath &&
-      workspaceSymbol.location.uri.fsPath.startsWith(workspacePath)
-    );
-  });
+
+  if (!symbols) return undefined;
+
+  const workspacePath = vscode.workspace.workspaceFolders?.[0].uri.path;
+
+  if (!workspacePath) return [];
+
+  return symbols
+    .map((workspaceSymbol) => {
+      const symbolPath = workspaceSymbol.location.uri.fsPath;
+      const relativePath = symbolPath.startsWith(workspacePath)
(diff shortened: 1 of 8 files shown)


yonip23:
i ran the chat in debug, and mocked hard-coded commands in the client request to see the result:
image

src/tabnineChatWidget/handlers/resolveSymbols.ts
```diff
@@ -1,18 +1,45 @@
...
+  if (!symbols) return undefined;
+
+  const workspacePath = vscode.workspace.workspaceFolders?.[0].uri.path;
+
+  if (!workspacePath) return [];
+
+  return symbols
+    .map((workspaceSymbol) => {
+      const symbolPath = workspaceSymbol.location.uri.fsPath;
+      const relativePath = symbolPath.startsWith(workspacePath)
```
dimacodota, line 25: use https://nodejs.org/api/path.html#pathrelativefrom-to

src/tabnineChatWidget/workspaceCommands/commandExecutors/findSymbols.ts
```diff
@@ -0,0 +1,13 @@
+import { CommandExecutor } from ".";
+import { resolveSymbols } from "../../handlers/resolveSymbols";
+
+const findSymbolsCommandExecutor: CommandExecutor = {
```
dimacodota, line 4: why the function declaration? just export the function

src/tabnineChatWidget/workspaceCommands/commandExecutors/index.ts
```diff
@@ -0,0 +1,3 @@
+export interface CommandExecutor {
```
dimacodota, line 1: can be T , instead of string[]

yonip23: i don't see a reason to make it generic, as eventually all the data must be `string[]` in order to get into the prompt, so this way I'm enforcing the executors to serialize it correctly

src/tabnineChatWidget/workspaceCommands/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  data: string[];
+};
+
+type CommandExecutor = (arg: string) => Promise<string[] | undefined>;
+
+const commandsExecutors: Record<WorkspaceCommand, CommandExecutor> = {
+  findSymbols: findSymbolsCommandExecutor,
+};
+
+export default async function executeWorkspaceCommand(
```
dimacodota, line 21: maybe you should handle errors here ? you are using Promise.all to map all the command executions.Promise all will fail it has a rejected promise. so either this will aways fulfill the promise without rejecting it or use `allSettled` instead of `all`

dimacodota (approved):
lgtm, see my last comment

src/tabnineChatWidget/workspaceCommands/index.ts
yonip23: good point - fixed using allSettled 👍🏻

src/tabnineChatWidget/handlers/resolveSymbols.ts
```diff
@@ -1,18 +1,46 @@
...
+      };
+    })
+    .filter(
+      (workspaceSymbol) => !!workspaceSymbol
+    ) as SymbolInformationResult[];
+}
+
+function isProbablyNotSource(symbolPath: string): boolean {
+  return ["node_modules", "dist", "build", "target", "out"].some((dir) =>
+    symbolPath.startsWith(dir)
```
dimacodota, line 44: r u  sure that in all cases the past will start with any of this?  i mean the relative path can start  with `/`  or  in java it could be a gazilion levels down

yonip23: So what r u suggesting? To check for contains instead of startswith?
Not sure it's a good idea either, I can imagine users naming some sub-folder "target" for example

PR #1195, opened 2023-07-09
DEV2-3070 add non existing handler

yonip23:


src/tabnineChatWidget/ChatViewProvider.ts
```diff
@@ -45,6 +45,12 @@ export default class ChatViewProvider implements WebviewViewProvider {
           });
         } catch (e) {
           console.error("failed to handle event. message:", message);
+          void this.chatWebview?.postMessage({
+            id: message.id,
+            payload: {
```
yairco1990, line 50: I think it will be better to return "error" instead of "payload". wdyt?

yairco1990: and this is how we handle it in the app
https://github.com/codota/tabnine-chat-app/compare/yair/handle_not_exists?expand=1

yairco1990 (approved):

dimacodota (approved):
lgtm

PR #1201, opened 2023-07-13
DEV2-2085: Log to output instead of dev tools

assapir:


src/utils/logger.ts
```diff
@@ -0,0 +1,26 @@
...
+  // eslint-disable-next-line no-use-before-define
+  private static instance: Logger;
+
+  private outputChannel: OutputChannel;
+
+  private constructor() {
+    this.outputChannel = vscode.window.createOutputChannel("Tabnine");
+  }
+
+  public static log(message: string): void {
```
dimacodota, line 14: consider to add a different log level output support (error/info)

assapir: with a different prefix?

dimacodota (approved):
lgtm 🚀

src/utils/logger.ts
assapir: ok, cool. will do

PR #1209, opened 2023-07-17
DEV2-3189 find symbols from user qurey in current file

yonip23:
1. if the user selected a code - use this selection.
2. else:
    1. divide the user query into words, and search each word as a symbol, using `vscode.executeWorkspaceSymbolProvider`.
    2. find the first symbol with the same URI as the current document, and use its code as the `selectedCode` field.
        this operation is cancelled if it takes more than 1s to complete, as we don't want to delay the response too much.


yonip23:
TODO: bump chat version before merging

src/utils/execution.utils.ts
```diff
@@ -0,0 +1,11 @@
+export async function executeTimeout<R>(
```
dimacodota, line 1: use rejectOnTimeout instead

src/tabnineChatWidget/handlers/getEditorContextHandler.ts
```diff
@@ -58,6 +67,58 @@ export async function getEditorContext(): Promise<EditorContextResponse> {
   };
 }
 
+async function getSelectedCode(
+  editor: vscode.TextEditor,
+  request?: EditorContextRequest
+): Promise<string | undefined> {
+  const selectedCode = editor.document.getText(editor.selection);
+  if ((selectedCode && selectedCode !== "") || !request?.userQuery) {
```
dimacodota, line 75: you can check on `editor.selection.isEmpty` instead

src/tabnineChatWidget/handlers/getEditorContextHandler.ts
```diff
@@ -58,6 +67,58 @@ export async function getEditorContext(): Promise<EditorContextResponse> {
...
+    return selectedCode;
+  }
+
+  const wordsInQuery = request.userQuery.split(/\s+/gm).slice(0, 20);
```
dimacodota, line 79: maybe something like 
``` 
const wordRegex = /\b\w+\b/g;
const matches = text.match(wordRegex);
```
would be more percise?

src/tabnineChatWidget/handlers/getEditorContextHandler.ts
```diff
@@ -58,6 +67,58 @@ export async function getEditorContext(): Promise<EditorContextResponse> {
...
+        )
+        .catch((e) => {
+          Logger.debug(
+            `failed to fetch symbols for '${word}': ${(e as Error).message}`
+          );
+          return undefined;
+        })
+    )
+  );
+  const firstSymbol = allSymbols.find((s) => s?.length)?.find((s) => !!s);
```
dimacodota, line 115: can you elaborate? why only the first symbol ?

src/tabnineChatWidget/handlers/getEditorContextHandler.ts
yonip23: which cases are you optimising with this regex? can you give an example of a prompt that requires this regex?

src/tabnineChatWidget/handlers/getEditorContextHandler.ts
yonip23: the selected code should be only one paragraph, so I figured lets go with the first occurrence of a symbol from the file.
What to you have in mind? to join all the occurrences?

dimacodota: yep, tough to join them all... but we can start only with the first one

dimacodota (approved):
lgtm

PR #1217, opened 2023-07-19
DEV2-2085 logs tooling

dimacodota:
new `tabnine.logEngine` setting introduced, by enabling  it the binary  stderr will be outputted to the extension log

Screen Shot 2023-07-19 at 23 45 40


assapir (approved):

src/utils/logger.ts
```diff
@@ -11,12 +11,23 @@ enum LogLevel {
...
   constructor() {
     this.outputChannel = vscode.window.createOutputChannel("Tabnine");
+    this.showLogsDisposable = vscode.commands.registerCommand(
+      "tabnine.logs",
+      () => this.show()
+    );
+  }
+
+  init(context: vscode.ExtensionContext) {
+    context.subscriptions.push(this);
```
assapir, line 25: why is that needed?

src/utils/logger.ts
```diff
@@ -40,6 +51,10 @@ export const Logger = new (class Logger implements vscode.Disposable {
     this.log(LogLevel.DEBUG, message, optionalParams);
   }
 
+  process(message: string): void {
+    this.outputChannel.appendLine(`${message}\n`);
```
assapir, line 55: I think it's better to have a prefix or something. this might cause a lot of noise and we should be able to tell if that a binary log or the extension log

src/binary/runProcess.ts
```diff
@@ -29,6 +29,10 @@ export function runProcess(
   );
   const proc = args ? spawn(command, args, options) : spawn(command, options);
 
+  proc.stderr?.on("data", (data: string) => {
+    Logger.process(data.toString());
```
assapir, line 33: data is already string, why the need for `toString()`?

assapir (approved):
Great idea! I wonder if we can have a "zip logs and send to support" somehow.

src/utils/logger.ts
assapir: ok, I read about it. cool.

dimacodota:
> Great idea! I wonder if we can have a "zip logs and send to support" somehow.
yep, i taught to add a "profile extension"  command that will collect the logs

PR #1220, opened 2023-07-21
DEV2-3247 - Fix: Startup Issue in devcontainers

mikekosulin:
Extension was failing to initialize correctly when launched in [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers). Upon investigating, it appears that users needed to manually input certain settings to workaround this issue.

To enable the Tabnine extension to work with devcontainers, users [have been required to add the following settings](https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location) to their VSCode settings.json:

```jsonc
// VSCode user's settings.json
{
  "remote.extensionKind": {
    "TabNine.tabnine-vscode": [
      "ui",
      "workspace"
    ]
  }
}

```

Considering the complexity and inconvenience this imposes on users, I've prepared this PR to address and resolve the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,10 @@
   "activationEvents": [
     "*"
   ],
+  "extensionKind": [
+    "ui",
+    "workspace"
+  ],
   "main": "./out/extension",
   "scripts": {
     "analyze:bundle": "webpack --env analyzeBundle",


package.json
```diff
@@ -93,6 +93,10 @@
   "activationEvents": [
     "*"
   ],
+  "extensionKind": [
+    "ui",
```
dimacodota, line 97: hey @mikekosulin,  is there any specific reason for choosing this order and not ["workspace", "ui"]?

mikekosulin: Yeah, I’ve checked a few options and ‘["workspace", "ui"]‘ doesn’t affect on the issue. So Tabnine extension activates if setting is ‘["ui", "workspace"]‘ or ‘["ui"]’ 

You can easily check it by following these steps to reproduce:
1. Start any devcontainer
2. Install Tabnine extension 
3. Add the the code below to your settings and reload the VS Code window:
‘’’
// VSCode user's settings.json
{
  "remote.extensionKind": {
    "TabNine.tabnine-vscode": [
      "workspace",
      "ui"
    ]
  }
}
‘’’

dimacodota (approved):
LGTM

amircodota:
@mikekosulin 

Hi, Mike

I am afraid we have to revert this PR.

It causes issues to many users working via remote ssh and other remote options.

I think we need to come up with a different solution for devcontainers - one that allows tabnine to run inside the container.

We will be looking into this and would love your input. So feel free to open an issue or contact our support and describe your situation in more detail

PR #1234, opened 2023-08-06
DEV2-3360: Add access to chat interface

ofekby:


package.json
```diff
@@ -221,6 +221,18 @@
...
+        "title": "Submit message"
+      },
+      {
+        "command": "tabnine.chat.state",
+        "title": "Chat global state"
+      },
+      {
+        "command": "tabnine.chat.clear-all-conversations",
+        "title": "Clear all tabnine chat conversations"
+      },
```
yonip23, line 235: Isn't this available to the users in the search bar as well?
maybe add a condition in the "when" field, and set it to true only when the eval is running?

yonip23 (commented):
looks good, see my comment

package.json
dimacodota: agree with @yonip23, another thing - the vscod's practice of exposing apis, is to return an object from the activate function, see an example usage here https://github.com/codota/tabnine-vscode/blob/b4e19e26e2cbd85bfa16aff075975e00d1b13df5/src/languages/java/JavaHome.ts#L14

dimacodota (approved):
lgtm, see comment

package.json
yonip23: im not sure i undestand what is the change you want here @dimacodota

ofekby: @dimacodota Chose to avoid refactoring all the code to expose these methods from the activate function, Used when...

PR #1240, opened 2023-08-08
DEV2-3414 bump version on alpha release

dimacodota:
DEV2-3414 bump version on alpha release


amircodota (approved):
LGTM

I suggest you write something to the entire team, explaining this change and what to expect in terms of alpha versions from now on

.github/workflows/alpha_release.yml
```diff
@@ -4,94 +4,105 @@ on:
...
-      id: delete-prerelease
-      uses: codota/delete-tag-and-release@master
-      with:
-        delete_release: true
-        tag_name: ${{ steps.prerelease-tag.outputs.release }}
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: bump version
+        run: |
+          yarn run apply-version ${{ github.run_id }}
```
amircodota, line 27: I think you can use run_number which would make this number smaller and more readable

amircodota (approved):
LGTM, but see my comment about run_number

.github/workflows/alpha_release.yml
dimacodota: the run_number is relative to the run an execution of the flow, no?

PR #1262, opened 2023-08-17
DEV2-3283: Add chat by user login state

ofekby:


src/enterprise/requests/UserInfo.ts
```diff
@@ -1,11 +1,15 @@
 import { tabNineProcess } from "../../binary/requests/requests";
 
+type KnownScopes = "chat" | "completions";
+
 export type UserInfo = {
   email: string;
   team: [];
   verified: [];
   isLoggedIn: boolean;
+  scopes?: KnownScopes[];
```
amirbilu, line 10: It's not optional in binary if I remember correctly

ofekby: It is for backward compitability

PR #1276, opened 2023-09-04
DEV2-3537: Remove all other extensions that user might have

assapir:
This PR makes the updater/self hosted extension which is `tabnine-vscode-self-hosted-updater` to remove both GA and the now deprecated enterprise extnsion

### DIFF
--- a/src/enterprise/extension.ts
+++ b/src/enterprise/extension.ts
@@ -43,10 +43,10 @@ export async function activate(
   initReporter(new LogReporter());
   context.subscriptions.push(new StatusBar(context));
 
-  void uninstallGATabnineIfPresent();
+  void uninstallAllOtherExtensionsIfPresent();
   context.subscriptions.push(
     vscode.extensions.onDidChange(() => {
-      void uninstallGATabnineIfPresent();
+      void uninstallAllOtherExtensionsIfPresent();
     })
   );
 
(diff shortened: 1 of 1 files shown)


src/enterprise/extension.ts
```diff
@@ -132,20 +132,31 @@ function registerAuthenticationProviders(
...
     const uninstall = await confirm(
       "⚠️ You have a conflicting version of Tabnine!",
       "Fix"
     );
     // the user provided consent
     if (uninstall) {
-      await commands.executeCommand(
-        "workbench.extensions.uninstallExtension",
-        "tabnine.tabnine-vscode"
+      await Promise.all(
```
yanir-codota, line 153: if it fails it will retry again next time? when is this triggered?
also, promiseAll rejects on first error. is this what you need?

assapir: it will happen again on the next startup, yes.
I don't care if it will be rejected, as it will try again 🤷🏻

dimacodota, line 153: did you tested it? promise all will fail if any the  promises is rejected,  now i'm not sure how `workbench.extensions.uninstallExtension` is behaves when a non existing extension name  is sent, but id change the `Promise.all` to `Promise.allSettled` and check if any is fulfilled

dimacodota (approved):
lgtm

PR #1293, opened 2023-09-14
DEV2-3664: active-text-editor

yonidavidson:
Notify binary on active text editor change

### DIFF
--- a/src/activeTextEditorState.ts
+++ b/src/activeTextEditorState.ts
@@ -1,0 +1,11 @@
+import * as vscode from "vscode";
+import { prefetch } from "./binary/requests/prefetch";
+
+export const activeTextEditorState = vscode.window.onDidChangeActiveTextEditor(
+  async (editor) => {
+    if (editor) {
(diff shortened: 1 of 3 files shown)


yonip23 (approved):
easy ✌️

src/activeTextEditorState.ts
```diff
@@ -0,0 +1,12 @@
+import * as vscode from "vscode";
+import { prefetch } from "./binary/requests/prefetch";
+
+// Notify the binary on changes to the active text editor.
+export const activeTextEditorState = vscode.window.onDidChangeActiveTextEditor(
```
dimacodota, line 5: if i remember correctly this will be fired on any scheme(git diff etc),  can you check this? an if so filter only the file scheme :
```
document.uri.scheme === "file"
```

dimacodota (approved):
lgtm

src/activeTextEditorState.ts
yonidavidson: validated and it gets the file name.

PR #1301, opened 2023-09-27
DEV2-3822: Copy disable proxy setting from updater

assapir:
if the user disabled the proxy setting in the updater, we should use this setting

### DIFF
--- a/src/enterprise/consts.ts
+++ b/src/enterprise/consts.ts
@@ -1,4 +1,10 @@
 export const SELF_HOSTED_SERVER_CONFIGURATION = `tabnineSelfHostedUpdater.serverUrl`;
+
+export const SELF_HOSTED_IGNORE_PROXY_CONFIGURATION =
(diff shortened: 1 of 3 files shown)


src/enterprise/consts.ts
```diff
@@ -1,4 +1,10 @@
 export const SELF_HOSTED_SERVER_CONFIGURATION = `tabnineSelfHostedUpdater.serverUrl`;
+
+export const SELF_HOSTED_IGNORE_PROXY_CONFIGURATION =
+  "tabnineSelfHostedUpdater.useProxySupport";
+
+export const IGNORE_PROXY_CONFIGURATION = "tabnine.useProxySupport";
```
dimacodota, line 6: it should be part of the package.json

assapir: it's not new, just in a const now because it is in use in few places

yanir-codota (approved):
LGTM

PR #1304, opened 2023-09-28
DEV2-3412: Change to use prettier for the whole repository

ofekby:


yanir-codota (approved):
sorry for the latency

PR #1307, opened 2023-10-02
DEV2-3856: Add command to get the server URL

assapir:
For now, it will return `https://api.tabnine.com` as default.  in the future it will ask the binary for that


amircodota (approved):
LGTM

PR #1345, opened 2023-10-29
DEV2-3868 sync status bar

amirbilu:


src/enterprise/statusBar/StatusBar.ts
```diff
@@ -66,7 +66,7 @@ export class StatusBar implements Disposable {
     this.item.setCommand(StatusState.WaitingForProcess);
 
     rejectOnTimeout(tabNineProcess.onReady, 10_000).then(
-      () => this.setLoginRequired(),
+      () => this.enforceLogin(),
```
yanir-codota, line 69: can this be called before actual login? I think the user has to be logged in to getUserInfo, is it promised that this already happened?

dimacodota (approved):
lgtm

src/enterprise/statusBar/StatusBar.ts
dimacodota: if the user is not logged in `getUserInfo` will return `isLoggedIn: false` this will set the "login required" state

PR #1349, opened 2023-11-01
DEV2-4002:  calculate workspace language fallback

yonidavidson:
Calculate workspace language based on extension of files.
This is used for cases chat is addressed but no language is detected.


src/tabnineChatWidget/handlers/context/basicContextHandler.ts
```diff
@@ -36,7 +36,76 @@ async function noEditorResponse() {
...
+
+  const topFileTypes = Array.from(fileTypeToLanguageID.keys());
+
+  // Ensure a workspace is open
+  if (!vscode.workspace.workspaceFolders) {
+    return;
+  }
+
+  // Get up to maxFiles from the workspace without depth restriction
+  const files = await vscode.workspace.findFiles("**/*", null, maxFiles);
```
dimacodota, line 85: few notes, 
1 - we already have a lang to ext mapping [here](https://github.com/codota/tabnine-vscode/blob/cc74eeec67703c2165071d51176cfaa1a98b7c55/src/globals/languages.ts#L1) 
2- you cab use workspace.textDocuments in order to get the correct languageId without the ext<>language mapping.

then the code could be something like:
```
function getMostUsedLanguage() {
  const languageCounts: { [key: string]: number } = {};

  workspace.textDocuments.forEach((document) => {
    const language = document.languageId;

    if (languageCounts[language]) {
      languageCounts[language] += 1;
    } else {
      languageCounts[language] = 1;
    }
  });

  const mostUsedLanguage = Object.keys(languageCounts).reduce((a, b) =>
    languageCounts[a] > languageCounts[b] ? a : b
  );
  return mostUsedLanguage;
}

dimacodota: + add tests for it

yonidavidson: Tests added, the workspace.textDocuments doesn't work when all files are closed.

dimacodota (approved):
lgtm

PR #1350, opened 2023-11-01
DEV2-4120 add "navigate_to_location" handler

yonip23:
https://github.com/codota/tabnine-vscode/assets/67855609/fd98b657-cabf-4a17-85a6-96929db2589c

### DIFF
--- a/src/tabnineChatWidget/ChatApi.ts
+++ b/src/tabnineChatWidget/ChatApi.ts
@@ -26,6 +26,10 @@ import {
   SelectedCodeResponsePayload,
   getSelectedCode,
 } from "./handlers/context/editorContext";
+import {
+  NavigateToLocationPayload,
+  navigateToLocation,
+} from "./handlers/navigateToLocation";
 
 type GetUserResponse = {
   token: string;
@@ -147,6 +151,10 @@ export function initChatApi(
     .registerEvent<{ symbols: vscode.SymbolInformation[] }, void>(
       "peek_definition",
       peekDefinition
+    )
+    .registerEvent<NavigateToLocationPayload, void>(
+      "navigate_to_location",
+      navigateToLocation
     )
     .registerEvent<ChatConversation, void>(
       "update_chat_conversation",
--- a/src/tabnineChatWidget/handlers/navigateToLocation.ts
+++ b/src/tabnineChatWidget/handlers/navigateToLocation.ts
@@ -17,11 +17,7 @@ async function doNavigate({
 }: NavigateToLocationPayload): Promise<void> {
   const uri = vscode.Uri.file(path);
   const document = await vscode.workspace.openTextDocument(uri);
-  await vscode.window.showTextDocument(document);
-  const editor = vscode.window.visibleTextEditors.find(
-    (visibleEditor) => visibleEditor.document === document
-  );
-  if (!editor) return;
+  const editor = await vscode.window.showTextDocument(document);
 
   const endLineLength = document.lineAt(range.endLine).text.length;
   const navigationRange = new vscode.Range(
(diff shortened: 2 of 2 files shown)


src/tabnineChatWidget/handlers/navigateToLocation.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+async function doNavigate({
+  path,
+  range,
+}: NavigateToLocationPayload): Promise<void> {
+  const uri = vscode.Uri.file(path);
+  const document = await vscode.workspace.openTextDocument(uri);
+  await vscode.window.showTextDocument(document);
+  const editor = vscode.window.visibleTextEditors.find(
```
dimacodota, line 21: why not use the editor  returned  from `showTextDocument`?

src/tabnineChatWidget/handlers/navigateToLocation.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export interface NavigateToLocationPayload {
+  path: string;
+  range: {
+    startLine: number;
+    endLine: number;
+  };
+}
+export function navigateToLocation(payload: NavigateToLocationPayload): void {
+  void doNavigate(payload);
```
dimacodota, line 11: why do you need the  wrapping function?

src/tabnineChatWidget/handlers/navigateToLocation.ts
```diff
@@ -0,0 +1,38 @@
...
+}: NavigateToLocationPayload): Promise<void> {
+  const uri = vscode.Uri.file(path);
+  const document = await vscode.workspace.openTextDocument(uri);
+  await vscode.window.showTextDocument(document);
+  const editor = vscode.window.visibleTextEditors.find(
+    (visibleEditor) => visibleEditor.document === document
+  );
+  if (!editor) return;
+
+  const endLineLength = document.lineAt(range.endLine).text.length;
```
dimacodota, line 26: it is minor, but if you need the whole last line, then why not just add 1 to the endLine and  start from zero position at the (endLine+ 1) line? 
then calculate the range as following:
```
const navigationRange = new vscode.Range(
    range.startLine,
    0,
    range.endLine+ 1,
    0
  );

File: src/tabnineChatWidget/handlers/navigateToLocation.ts (hunk 2 above)
Comment (yonip23) [reply src/tabnineChatWidget/handlers/navigateToLocation.ts:11]: to have multiple awaits and `void` them outside, otherwise i'd have to use `.then`s

File: src/tabnineChatWidget/handlers/navigateToLocation.ts (hunk 3 above)
Comment (yonip23) [reply src/tabnineChatWidget/handlers/navigateToLocation.ts:26]: in IJ i implemented it this way, and would like to do the same here for consistency.

File: src/tabnineChatWidget/handlers/navigateToLocation.ts (hunk 2 above)
Comment (dimacodota) [reply src/tabnineChatWidget/handlers/navigateToLocation.ts:11]: And why do you need to void on them ?

File: src/tabnineChatWidget/ChatApi.ts [hunk 4]
```diff
@@ -148,6 +152,10 @@ export function initChatApi(
       "peek_definition",
       peekDefinition
     )
+    .registerEvent<NavigateToLocationPayload, void>(
+      "navigate_to_location",
+      navigateToLocation
```
dimacodota, line 157: and if you do ned to avoid on it, it would be better to do it here

src/tabnineChatWidget/handlers/navigateToLocation.ts
yonip23: oh, im just now seeing that the chat API registry supports async handlers as well, will change

src/tabnineChatWidget/ChatApi.ts
yonip23: what?

src/tabnineChatWidget/handlers/navigateToLocation.ts
dimacodota: nvm, you've changed the navigateToLocation

dimacodota (approved):
lgtm

PR #1365, opened 2023-11-16
DEV2-4225 support multi workspace in retrieval-related features

yonip23:
### Indexing 

https://github.com/codota/tabnine-vscode/assets/67855609/a0982345-3790-4899-bbcd-83d9821cc01c

### Querying
Screenshot from 2023-11-16 14-41-18
Screenshot from 2023-11-16 14-41-33

### DIFF
--- a/src/WorkspaceUpdater.ts
+++ b/src/WorkspaceUpdater.ts
@@ -2,6 +2,7 @@ import * as vscode from "vscode";
 import { BINARY_UPDATE_WORKSPACE_INTERVAL } from "./globals/consts";
 import sendUpdateWorkspaceRequest from "./binary/requests/workspace";
 import { Logger } from "./utils/logger";
+import { tabNineProcess } from "./binary/requests/requests";
 
 const INITIAL_DELAY = 3000;
 
@@ -9,13 +10,15 @@ export class WorkspaceUpdater implements vscode.Disposable {
   private updateInterval: NodeJS.Timeout | null = null;
 
   constructor() {
-    setTimeout(() => {
-      updateWorkspace();
-      this.updateInterval = setInterval(
-        updateWorkspace,
-        BINARY_UPDATE_WORKSPACE_INTERVAL
-      );
-    }, INITIAL_DELAY);
+    void tabNineProcess.onReady.then(() => {
+      setTimeout(() => {
+        updateWorkspace();
+        this.updateInterval = setInterval(
+          updateWorkspace,
+          BINARY_UPDATE_WORKSPACE_INTERVAL
+        );
+      }, INITIAL_DELAY);
+    });
   }
 
   dispose() {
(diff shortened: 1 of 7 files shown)


src/enterprise/extension.ts
```diff
@@ -133,6 +137,7 @@ function initSelectionHandling(): vscode.Disposable {
 }
 
 export async function deactivate(): Promise<unknown> {
+  cancelWorkspaceUpdater();
   return requestDeactivate();
```
amircodota, line 141: Maybe this should go inside `requestDeactivate`?
And maybe there's also some common code in activate between enterprise and saas?

amircodota (approved):
LGTM. Had some minor comments?

src/enterprise/extension.ts
yonip23: > Maybe this should go inside `requestDeactivate`?

This API only deactivates the binary process. Other intervals are canceled outside this function.

>  And maybe there's also some common code in activate between enterprise and saas?

Unless @dimacodota knows something else, I didn't find such common code (maybe worth doing it, but not in this pr)

dimacodota: id suggest to convert the startWorkspaceUpdater  to a WorkspaceUpdater that implements `Disposable `
and then add it to `context.subscriptions` this will handle all the cleanup 

see https://github.com/codota/tabnine-vscode/blob/77451b9438459ed8db2e56e13a40e3d9dde432e9/src/enterprise/statusBar/StatusBar.ts#L16 fotr exmple

src/workspaceUpdater.ts
```diff
@@ -0,0 +1,45 @@
...
+import { BINARY_UPDATE_WORKSPACE_INTERVAL } from "./globals/consts";
+import sendUpdateWorkspaceRequest from "./binary/requests/workspace";
+import { Logger } from "./utils/logger";
+
+const INITIAL_DELAY = 3000;
+
+let updateInterval: NodeJS.Timeout | null = null;
+
+export default function startWorkspaceUpdater() {
+  setTimeout(() => {
```
dimacodota, line 11: you need to start the loop after the binary is ready, see https://github.com/codota/tabnine-vscode/blob/77451b9438459ed8db2e56e13a40e3d9dde432e9/src/authentication/TabnineAuthenticationProvider.ts#L89 for exmple

src/WorkspaceUpdater.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  dispose() {
+    if (this.updateInterval) {
+      clearInterval(this.updateInterval);
+    }
+  }
+}
+
+function updateWorkspace() {
+  const rootPaths = vscode.workspace.workspaceFolders
```
dimacodota, line 32: pls  check that it  returns all the workspace and only the opened workspaces 

from the doc 
```
List of workspace folders (0-N) that are open in the editor. undefined when no workspace has been opened.

dimacodota (approved):
lgtm

src/WorkspaceUpdater.ts
yonip23: yes, I checked

PR #1376, opened 2023-11-26
DEV2-4157: Add tabnine chat state

ofekby:


src/utils/EventEmitterBasedState.ts
```diff
@@ -0,0 +1,46 @@
+import { Mutex } from "await-semaphore";
+import { isEqual } from "underscore";
+import { Disposable, EventEmitter } from "vscode";
+
+export default class EventEmitterBasedState<T> {
```
dimacodota, line 5: i think it should be disposable bc EventEmitter is disposable.

moreover EventEmitterBasedState can extend  EventEmitter  instead of composition

src/utils/EventEmitterBasedNonNullState.ts
```diff
@@ -0,0 +1,15 @@
+import EventEmitterBasedState from "./EventEmitterBasedState";
+
+export default class EventEmitterBasedNonNullState<
```
dimacodota, line 3: why do you need this wrapper? in what case null state is valid?

src/utils/deriveState.ts
```diff
@@ -0,0 +1,45 @@
+import { Disposable } from "vscode";
+import EventEmitterBasedState from "./EventEmitterBasedState";
+
+export type DerivedState<T> = Disposable & EventEmitterBasedState<T>;
+
+export default function deriveState<I, O, S extends EventEmitterBasedState<I>>(
```
dimacodota, line 6: why is it exported?

dimacodota:
lets move all the "state management" under state folder instead of utils

dimacodota (approved):
lgtm, has some minor comments

src/utils/EventEmitterBasedNonNullState.ts
ofekby: Mostly for typescript. Null is the default initial state. But this wrapper is for when you an actual initial state. By the current state design, the value is only null able before initialisation after that the state will never change back to null. And the listeners are only notified on when the non null value has changed.

PR #1380, opened 2023-11-27
DEV2-4344 add workspace trace

amircodota:


src/WorkspaceUpdater.ts
```diff
@@ -28,11 +29,30 @@ export class WorkspaceUpdater implements vscode.Disposable {
...
 }
 
+async function sendEvent(event: string, payload: Record<string, unknown> = {}) {
+  if (isCapabilityEnabled(Capability.WORKSPACE_TRACE)) {
+    fireEvent({
+      name: event,
+      ...payload,
+    });
+  }
+}
```
yonip23, line 39: you can add the `void` here and get rid of it in the callers
```suggestion
function sendEvent(event: string, payload: Record<string, unknown> = {}) {
  if (isCapabilityEnabled(Capability.WORKSPACE_TRACE)) {
    void fireEvent({
      name: event,
      ...payload,
    });
  }
}
```

yonip23 (approved):
see comment

PR #1383, opened 2023-11-28
DEV2-4342 log file and level from env

yonip23:
for the e2e eval - i need to see the binary logs to debug what's going on there, so the e2e eval will pass these env args and upload the log file

### DIFF
--- a/src/globals/tabnineExtensionProperties.ts
+++ b/src/globals/tabnineExtensionProperties.ts
@@ -55,10 +55,10 @@ function getContext(): TabNineExtensionProperties {
   const autoImportConfig = "tabnine.experimentalAutoImports";
   const logFilePath =
     configuration.get<string>("tabnine.logFilePath") ??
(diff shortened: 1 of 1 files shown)


src/globals/tabnineExtensionProperties.ts
```diff
@@ -53,8 +53,12 @@ function getContext(): TabNineExtensionProperties {
     "typescript.suggest.autoImports"
   );
   const autoImportConfig = "tabnine.experimentalAutoImports";
-  const logFilePath = configuration.get<string>("tabnine.logFilePath");
-  const logLevel = configuration.get<string>("tabnine.logLevel");
+  const logFilePath =
+    configuration.get<string>("tabnine.logFilePath") ??
+    process.env.BINARY_LOG_FILE_PATH;
```
amircodota, line 58: maybe name the env variables `TABNINE_...`?

amircodota (approved):
LGTM but see my comment regarding naming

PR #1404, opened 2023-12-10
DEV2-4455 - check if intellicode is running to avoid collision

dimacodota:
check if intellicode is running to avoid collision


amircodota (approved):
LGTM

PR #1409, opened 2023-12-13
DEV2-564 Enhance context gathering in Jupyter notebooks by including text from…

taltabnine:
… all cells

### DIFF
--- a/src/runCompletion.ts
+++ b/src/runCompletion.ts
@@ -1,4 +1,5 @@
 import { CancellationToken, Position, Range, TextDocument } from "vscode";
+import * as vscode from "vscode";
 import {
   autocomplete,
   AutocompleteParams,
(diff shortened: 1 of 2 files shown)


src/runCompletion.ts
```diff
@@ -1,4 +1,5 @@
 import { CancellationToken, Position, Range, TextDocument } from "vscode";
+import * as vscode from "vscode";
```
dimacodota, line 2: there is a named import from vscode one line above, just add the `window` and the `Uri` to it

src/runCompletion.ts
```diff
@@ -51,6 +76,16 @@ export default async function runCompletion({
...
     sdk_path: getSDKPath(document.languageId),
   };
+  const notebookEditor = vscode.window.activeNotebookEditor;
+
+  if (notebookEditor) {
+    const { before, after } = calculateContextForJupyterNotebook(
+      notebookEditor,
+      document.uri
+    );
+    requestData.before = [before, requestData.before].join("\n");
```
dimacodota, line 86: i'd prefer an approach that's more "immutable". how about introducing a `getContextParts` function that returns something like `{before: string, after: string}`? this function could calculate the 'before' and 'after' based on whether it's a notebook or a regular document, instead of directly mutating the requestData. WDYT?

